PR #1, opened 2022-09-12
Assignments/boolean birds

elijahsur:


assignments/boolean-birds/code.js
```diff
@@ -15,4 +15,18 @@
  *   - Flibble Birds eat fish, mice, and worms.
  *   - Globby Birds eat fish, mice, and nuts.
  *
- */
+ */
```
gigamonkey, line 18: I looked through the history of the changes from the beginning to this. It was quite a journey of discovery. Which is actually great. I'm sure it was painful but hopefully having gotten there the hard way some of the ideas have sunk in. Anyway, nice job persevering.

assignments/boolean-birds/code.js
```diff
@@ -15,4 +15,18 @@
  *   - Flibble Birds eat fish, mice, and worms.
  *   - Globby Birds eat fish, mice, and nuts.
  *
- */
+ */ 
+const isFlobbyBird = (a,b) => a&&b
```
gigamonkey, line 19: This looks great. You can reformat this code by right clicking in the editor pane and choosing "Format document" from the menu that pops up. It will add some spaces that will make this a bit more legible. Something like this: 

```suggestion
const isFlobbyBird = (a, b) => a && b;
```

assignments/boolean-birds/code.js
```diff
@@ -15,4 +15,18 @@
...
+const isBloggyBird = (a,b) => a&&!b
+const isFlibbleBird = (a,b) => !a&&b
+const isGlobbyBird = (a,b) => !a&&!b
+
+const eatsWorms = (a,b) => a&&b||!a&&b||a&&!b
```
gigamonkey, line 24: I feel like somewhere in the history of this function you actually hit on a simpler version:

```suggestion
const eatsWorms = (a, b) => a || b
```
But these versions have some merit in that you used the things you figured out from the other functions. To take that insight to the next level you could have used the functions you already wrote and done this:

```suggestion
const eatsWorms = (a, b) => isFlobbyBird(a, b) || isFlibbleBird(a, b) || isBloggyBird(a, b);
```

That becomes pretty readable. And at that point you might realize that that is the same as saying:

```suggestion
const eatsWorms = (a, b) => !isGlobbyBird(a, b);
```

I.e. It's not the one bird that doesn't eat worms. I didn't actually expect that. I expected the `a || b` solution. Another student came up with the `!isGlobbyBird(a, b)` style. which is pretty clever.

But I do look at your version and think there's gotta be a simpler way. And you can actually simplify boolean expressions kinda like you simplify mathematical expressions. So whenever you see

```
a && b || !a && b
```

you can simplify to:

```
b
```

because `a` will either be true or false so the truth value of `a && b || !a && b` will come down to just the value of `b`.

So if we simplify the first two expressions like that, then we have:

```
b || a && !b
```

And in that case we know that the result will only depend on the second part of the `||` expression when `b` is false (because if `b` was true the whole expression is true regardless of the value of the second expression.) So that can be simplified to:

```
b || a && !false
```

which is the same as

```
b || a && true
```

which is the same as:

```
b || a
```

So even without looking back at the details of the birds of Booleana Island, knowing that your implementation is correct I can simplify it with essentially algebra. Anyway, so you just got a big brain dump about boolean simplification. Hopefully it makes some sense. If not, no worries. We'll probably come back to it at some point in class.

assignments/boolean-birds/code.js
```diff
@@ -15,4 +15,18 @@
...
+const eatsNuts = (a,b) => a&&b||!a&&!b||a&&!b
+const eatsFish = (a,b) => a&&b||!a&&!b||!a&&b
+const eatsMice = (a,b) => a&&!b||!a&&!b||!a&&b
+
+const isRed = (a) => a === 'Bloggy' || a === 'Flobby'
```
gigamonkey, line 29: Took you a while to get here too and again I admire your sticktoitiveness. Hopefully you understand now that the argument to the function is a string and in order to turn a string into a boolean you need to compare it to something as you did here with '==='.

gigamonkey (approved):
This is fine and you can merge it straight away if you'd like. Though you might want to go through the exercise of simplifying the logic of the eats<whatever> functions.

Also, one other stylistic point is you could choose slightly more meaningful variable names. For instance `red` and `spotted` rather than `a` and `b`, and for the functions that take a single string maybe `bird` or `name` or something. Feel free to make those changes if you like in this code or just keep it in mind for future.

PR #2, opened 2022-09-12
Assignments/pig latin

elijahsur:


assignments/pig-latin/code.js
```diff
@@ -31,3 +31,5 @@
 //
 // Feel free to play around in the REPL to make sure you understand how this
 // operator works.
+const pigLatin = (word) => (word.substring(word.search(/[aeiou]/))+(word.substring(0,word.search(/[aeiou]/))+'ay'))
```
gigamonkey, line 34: Very nice over all. One style nit: you don't need the `()`s around the expression. So you can write it like this:

```suggestion
const pigLatin = (word) => word.substring(word.search(/[aeiou]/))+(word.substring(0,word.search(/[aeiou]/))+'ay');
```

One other idea is to introduce a variable to avoid having to search for the position of the first vowel twice. Saves a bit of work for the computer and makes it arguably more readable:

```javascript
const pigLatin = (word) => {
  const i = word.search(/[aeiou]/);
  return word.substring(i) + word.substring(0, i) + 'ay';
};
```

This is more readable for a couple reasons.

1. You can understand the first line on it's own: it's finding the index of the first vowel.
2. The final expression is easier to read just because it's shorter and also because it's easier  to see the `i` being used two places rather than having to recognize the much longer `word.search(/[aeiou]/)` expression is the same in both places.

gigamonkey (approved):
Looks good. Go ahead and merge!

PR #3, opened 2022-09-12
Assignments/number functions

elijahsur:


assignments/number-functions/code.js
```diff
@@ -0,0 +1,10 @@
+const add = (a,b) => (a+b);
```
gigamonkey, line 1: Overall looks good. Note you don't need `()`s around the expression. So this could just be.

```suggestion
const add = (a, b) =>  a + b;
```

And spaces around the operators and after commas is good style. If you right click in the editor pane you'll get a contextual menu that has an item "Format document". You can use that to fix a lot of these formatting issues.

assignments/number-functions/code.js
```diff
@@ -0,0 +1,10 @@
...
+const subtract = (a,b) => (a-b);
+const multiply = (a,b) => (a*b);
+const divide = (a,b) => (a/b);
+const mod = (a,b) => (a%b);
+const averageOfTwo = (a,b) => ((a+b)/2)
+const averageOfThree = (a,b,c) => ((a+b+c)/3)
+const distance = (a,b) => Math.abs(a-b)
+const manhattanDistance = (x1,y1,x2,y2) => (Math.abs(x1-x2)+Math.abs(y1-y2))
+const euclideanDistance = (x1,y1,x2,y2) => Math.sqrt(((Math.abs(x1-x2)**2)+(Math.abs(y1-y2)**2)))
```
gigamonkey, line 10: Just a mathematical note, You don`t need the `Math.abs` here since squaring a number will always result in a positive number. But this is fine.

gigamonkey (approved):
If you wanted you could run "Format document" on this just to keep things tidy but you can also go ahead and hit Merge as is.

PR #4, opened 2022-09-21
Assignments/images challenge

elijahsur:
still working on this, got one of them done

### DIFF
--- a/assignments/images-challenge/code.js
+++ b/assignments/images-challenge/code.js
@@ -96,4 +96,5 @@ const lineC = (count) => {
     yaxis = yaxis - 20
 }
 }
-lineC(20)
+
+drawLine(0,0,width,height,'black',1)
--- a/assignments/images-challenge/code.js
+++ b/assignments/images-challenge/code.js
@@ -32,8 +32,7 @@ const forlineOfCircles = (radius) => {
     drawFilledCircle(dif, height / 2, radius, 'red')
     dif = dif + radius * 2
   }
-}
-
+};
 const lineOfCircles = (radius) => {
   let dif = radius;
   let x = 0;
@@ -42,8 +41,7 @@ const lineOfCircles = (radius) => {
     x = x + 1
     dif = dif + radius * 2
   }
-}
-
+};
 const altLineOfCircles = (radius) => {
   let dif = radius;
   let count = 0;
@@ -78,5 +76,4 @@ const bigCircle = (count) => {
     alternate = alternate + 1
     factor = height / 2 - ((height/2) / count * c)
   }
-}
-bigCircle(40)
+};
--- a/assignments/images-challenge/code.js
+++ b/assignments/images-challenge/code.js
@@ -96,4 +96,4 @@ const lineC = (count) => {
     yaxis = yaxis - 20
 }
 }
-lineC(200)
+lineC(20)
--- a/assignments/images-challenge/code.js
+++ b/assignments/images-challenge/code.js
@@ -93,8 +93,6 @@ const lineC = (count) => {
   for (let c = 0; c < width/count + 1;c++) {
     drawLine(0, yaxis, xaxis, height, 'black', 1)
     xaxis = xaxis + 20
-    yaxis = yaxis - 20
+    yaxis = yaxis + 20
 }
 }
-
-drawLine(0,0,0,height,'black',1)
--- a/assignments/images-challenge/code.js
+++ b/assignments/images-challenge/code.js
@@ -97,4 +97,4 @@ const lineC = (count) => {
 }
 }
 
-drawLine(0,0,width,height,'black',1)
+drawLine(0,0,0,height,'black',1)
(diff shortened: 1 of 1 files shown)


assignments/images-challenge/code.js
```diff
@@ -23,3 +23,37 @@
...
+    drawFilledCircle(dif, height / 2, radius, color)
+    x = x + 1
+    dif = dif + radius * 2
+    if (alt % 2 == 0);
+    color = 'blue';
+    if (alt % 2 == 1);
+    color = 'red';
+    alt = alt + 1
+  }
+  return "done!"
```
gigamonkey, line 58: As I talked about today in class, when you're writing functions for their side effects (such as drawing on the screen) you don't really need to return anything. It doesn't really matter if you do but it's a bit distracting to someone reading the code who then has to think about why you're returning a value.

assignments/images-challenge/code.js
```diff
@@ -23,3 +23,37 @@
...
+  let dif = radius;
+  let x = 0;
+  let alt = 1;
+  let color = 'red'
+  while (x < Math.floor(width / (radius * 2))) {
+
+    drawFilledCircle(dif, height / 2, radius, color)
+    x = x + 1
+    dif = dif + radius * 2
+    if (alt % 2 == 0);
```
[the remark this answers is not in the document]
elijahsur: could you explain what the ";" characters do? i keep seeing them in code, but i don't fully understand their use

assignments/images-challenge/code.js
elijahsur: i mainly just used "return "done!" to stop the "undefined" answer from the REPL from showing, but i understand how it could mess with the reading of the code

assignments/images-challenge/code.js
```diff
@@ -23,3 +23,37 @@
  *
  *  clear()
  */
+
+
+
+const lineOfCircles = (radius) => {
+  let dif = radius;
+  let x = 0;
```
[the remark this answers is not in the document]
elijahsur: i'll look over the slide in the meantime, since i don't really remember much of the "for" loop

assignments/images-challenge/code.js
gigamonkey: Ah, that's actually a pretty okay reason. Eventually you should take it out but if you're using it for debugging while you're developing, that's cool.

assignments/images-challenge/code.js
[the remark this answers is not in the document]
gigamonkey: I think we talked about this a bit in class but basically they are used to delimit "statements". For now the kinds of statements you know about that should have a `;` at the end are assignments and returns. E.g.

```javascript
let x = 10;
return a;
y = 200;
```

An `if` statement is more complicated and technically has the form:

`if (`condition`)` 

followed by either a single statement (which will end with a `;`) or a block contained in `{}`

So either:

```javascript
if (whatever)
   return something;
```

Or

```javascript
if (whatever) {
  return something;
}
```

I'd *strongly* recommend *always* using `{}`s as doing so will prevent a whole bunch of common mistakes.

assignments/images-challenge/code.js
```diff
@@ -23,3 +23,51 @@
...
+    alternate = alternate + 1
+  }
+};
+const bigCircle = (count) => {
+  let color = 'red';
+  let alternate = 250;
+  let factor = width+height;
+  drawFilledCircle(width / 2, height / 2, 250, color)
+  for (let c = 0; c < count+1; c++) {
+    drawFilledCircle(width / 2, height / 2, factor, color)
```
gigamonkey, line 63: The first iteration of this loop it seems like this would draw a huge circle since `factor` starts out at `width + height`?

assignments/images-challenge/code.js
```diff
@@ -23,3 +23,51 @@
...
+    }
+    if (alternate % 2 == 1) {
+      color = 'red';
+    }
+    alternate = alternate + 1
+  }
+};
+const bigCircle = (count) => {
+  let color = 'red';
+  let alternate = 250;
```
gigamonkey, line 59: I'm a little confused what this variable is supposed to be. It starts at 250 and gets incremented by one but seems to be used only to detect whether you're on an odd or even circle. Feels redundant to me.

gigamonkey (changes requested):
I think you can simplify this code a bit and address the comments I left.

assignments/images-challenge/code.js
elijahsur: i was playing around with that width+height, trying to figure out a good replacement for my previous method of just using "250". the width+height is not permanent

assignments/images-challenge/code.js
elijahsur: i think i accidentally set "alternate" to 250 while trying to change the "factor" variable. it's meant to be at 0, which ill fix now

elijahsur:
done with this

gigamonkey (approved):
Sounds good. You can go ahead and merge this.

elijahsur:
cant seem to merge this, says it needs a review still

PR #5, opened 2022-09-30
Assessments/unit 01

elijahsur:


assessments/unit-01/code.js
```diff
@@ -5,3 +5,53 @@ const emit = (a, b, product) => {
...
+}
+
+const canSleepIn = (weekday, vacation) => {
+}
+
+const moreThanTwiceAsLong = (oneword,twoword) => {
+  if (oneword.length > twoword.length*2) {
+    return true
+  };
+  if (oneword.length < twoword.length*2) {
```
gigamonkey, line 23: This is really close. However you've made it too complicated and thus missed a case: if the first condition (`oneword.length > twoword.length * 2`) is not true you can just return false. But because you explicitly include the check you now miss the case where `oneword.length === twoword.length * 2`. As a consequence this function will return `undefined` in that case. Which arguably is close to correct because `undefined` counts as a "falsy" value in Javascript. But I'm pretty sure you weren't counting on that.

assessments/unit-01/code.js
```diff
@@ -5,3 +5,53 @@ const emit = (a, b, product) => {
...
+    return false
+  };
+}
+
+const firstHalf = (whole) => {
+  return whole.substring(0, whole.length / 2)
+}
+
+const secondHalf = (whole) => {
+  return whole.substring(whole.length / 2, whole.length)
```
gigamonkey, line 33: This is correct. Note you can also use the one-argument version: `whole.substring(whole.length / 2)` since the second argument defaults to the length of the string if not provided.

assessments/unit-01/code.js
```diff
@@ -5,3 +5,53 @@ const emit = (a, b, product) => {
...
+}
+
+const upDown = (whole) => {
+  return whole.toUpperCase() + whole.toLowerCase()
+}
+
+const everyOther = (whole) => {
+  return whole.substring(0, 1) + whole.substring(2, 3) + whole.substring(4, 5)
```
gigamonkey, line 41: This is correct but maybe even easier would be to use `[]`: i.e. `whole[0] + whole[2] + whole[4]`.

assessments/unit-01/code.js
```diff
@@ -5,3 +5,53 @@ const emit = (a, b, product) => {
...
+}
+
+const upDownLastCharacter = (whole) => {
+  return whole.substring(whole.length - 1, whole.length).toUpperCase() + whole.substring(whole.length - 1, whole.length).toLowerCase()
```
gigamonkey, line 45: Again correct but using both the two-arg version of `substring` and using `substring` instead of `[]` makes this a lot more code to read than the equivalent `whole[whole.length - 1].toUpperCase() + whole[whole.length - 1].toLowerCase()`.

And maybe even better would be to introduce a variable:

```javascript
const upDownLastCharacter = (whole) => {
  const last = whole[whole.length - 1];
  return last.toUpperCase() + last.toLowerCase();
};
```

assessments/unit-01/code.js
```diff
@@ -5,3 +5,53 @@ const emit = (a, b, product) => {
...
+}
+
+const yesIfEven = (number) => {
+  if (number % 2 == 0) {
+    return 'yes'
+  };
+  if (number % 2 == 1) {
```
gigamonkey, line 52: This one works because `== 0` and `== 1` are in fact all the possibilties. But the code would be even easier to read with an `if`/`else`:

```javacript
if (number % 2 === 0) {
  return 'yes';
} else {
  return 'no';
}
```

And even tidier, if you remember the `?:` operator from the pig latin assignment:

```javascript
return number % 2 === 0 ? 'yes" : 'no';
```

gigamonkey (changes requested):
Marking this as request changes since you should still do the remaining problems. But this looks pretty good. Some comments about specifics below.

assessments/unit-01/code.js
elijahsur: I just forgot the way to write [], so I kind of defaulted to using substrings

assessments/unit-01/code.js
elijahsur: hopefully fixed!

assessments/unit-01/code.js
elijahsur: fixed

assessments/unit-01/code.js
elijahsur: simplified!

assessments/unit-01/code.js
elijahsur: simplified!

PR #6, opened 2022-10-06
Assignments/number function practice

elijahsur:


assignments/number-function-practice/code.js
```diff
@@ -0,0 +1,66 @@
...
+}
+const remainder = (a, b) => {
+  return a % b
+}
+const average = (a, b) => {
+  return (a + b) / 2
+}
+
+const twoDistance = (x1, x2, y1, y2) => {
+  return Math.sqrt(Math.abs(x1 - x2) * Math.abs(x1 - x2) + Math.abs(y1 - y2) * Math.abs(y1 - y2))
```
gigamonkey, line 21: This is correct but you could make it more concise if you used the `**` exponentiation operator. Also you don't actually need the `Math.abs` since you're squaring the difference which will ensure that it's positive.

```suggestion
  return Math.sqrt((x1 - x2) ** 2 + (y1 - y2) ** 2)
```

assignments/number-function-practice/code.js
```diff
@@ -0,0 +1,66 @@
...
+const areaTriangle = (base, height) => {
+  return base / 2 * height
+}
+
+const baseTrapezoid = (b1, b2, height) => {
+  return (b1 + b2) / 2 * height
+}
+
+const areaCircle = (r) => {
+  return Math.pi * (r ** 2)
```
gigamonkey, line 53: Close but it's `Math.PI` with an upper case `PI`.

gigamonkey (approved):
Overall looks good. Seems like you should be ready for the assessment.

assignments/number-function-practice/code.js
elijahsur: i mainly didn't use the ** operator since for some reason it hasn't worked for me at some times before (100% human error). i also just threw in the Math.abs just to make sure.

PR #7, opened 2022-10-07
Assessments/unit 01 numeric

elijahsur:
didn't know any way to do the "volumeOfCube" one, i haven't used an edge to find a volume of a cube before (to my memory)

### DIFF
--- a/assessments/unit-01-numeric/code.js
+++ b/assessments/unit-01-numeric/code.js
@@ -40,6 +40,8 @@ const firstClassPostage = (weight) => {
   return 60+Math.ceil(weight-1)*24
 }
 
-const weightOnJupiter = (weightearth) => {}
+const weightOnJupiter = (weightearth) => {
+  return weightearth*(EARTH_GRAVITY/JUPITER_GRAVITY)
+}
 
 const gravity = (mass1, mass2, distance) => {}
--- a/assessments/unit-01-numeric/code.js
+++ b/assessments/unit-01-numeric/code.js
@@ -7,3 +7,37 @@ const G = 6.6743e-11;
 // syntactically correct function skeleton, more credit for a skeleton with a
 // reasonable argument list, yet more credit for a working function, and full
 // credit for a clear and simple working function.
+
+const itemsLeftOver = (npeople, nitems) => {
+  return nitems%npeople
+}
+
+const areaOfCircle = (radius) => {
+  return Math.PI*(radius**2)
+}
+
+const volumeOfCube = (length) => {
+  return 
+}
+
+const populationGrowth = (population, growthrate) => {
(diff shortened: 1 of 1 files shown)


assessments/unit-01-numeric/code.js
```diff
@@ -7,3 +7,46 @@ const G = 6.6743e-11;
...
+const itemsLeftOver = (npeople, nitems) => {
+  return nitems%npeople
+}
+
+const areaOfCircle = (radius) => {
+  return Math.PI*(radius**2)
+}
+
+const volumeOfCube = (length) => {
+  return
```
gigamonkey, line 20: I assume the problem here was mathematical not programming. The volume of a cube, for future reference, is the cube of the length of an edge. So in this case `length ** 3`. (This is analogous to how the area of a square is the square of the length of one side of the square.)

assessments/unit-01-numeric/code.js
```diff
@@ -7,3 +7,46 @@ const G = 6.6743e-11;
...
+}
+
+const valueOfJewels = (ndiamonds, nemeralds, vdiamond, vemerald) => {
+  return ndiamonds*vdiamond+nemeralds*vemerald
+}
+
+const payWithOvertime = (nhours, hourlyrate, overtime) => {
+  if (nhours > 8)
+    return (nhours-8)*overtime+8*hourlyrate
+    else
```
gigamonkey, line 38: Formatting is a bit wonky here. The `else` should be indented the same amount as the `if`. And I strongly, strongly urge you to always use `{}`s around the body of `if`, `else`, `for`, and `while` constructs. In this case that would look like:

```javascript
  if (nhours > 8) {
    return (nhours-8)*overtime+8*hourlyrate
  } else {
    return nhours*hourlyrate
  }
```

assessments/unit-01-numeric/code.js
```diff
@@ -7,3 +7,46 @@ const G = 6.6743e-11;
...
+    return nhours*hourlyrate
+}
+
+const firstClassPostage = (weight) => {
+  return 60+Math.ceil(weight-1)*24
+}
+
+const weightOnJupiter = (weightearth) => {
+  return weightearth*(EARTH_GRAVITY/JUPITER_GRAVITY)
```
gigamonkey, line 47: This is very close but you used the ratio of Earth's gravity to Jupiter's rather than the ratio of Jupiter's to Earth's. I.e. this should be `weightearth * (JUPITER_GRAVITY / EARTH_GRAVITY)`.

gigamonkey (approved):
Overall a very good job! Just a small mathematical mistake and forgetting (or maybe never having learned? if so, sorry about that) one mathematical fact. You can merge this.

PR #9, opened 2022-10-17
Assessments/unit 01 booleans and strings

elijahsur:
i forgot how to do the booleans in a function style, and i spent a good half of the time working on the other assignment, since i thought i didn't have to do the assessment

### DIFF
--- a/assessments/unit-01-booleans-and-strings/code.js
+++ b/assessments/unit-01-booleans-and-strings/code.js
@@ -31,5 +31,5 @@ const swapFrontAndBack = (string) => {
 }
 
 const simplePigLatin = (string,number) => {
-  return string.substring(number,string.length)+string.substring(0,number)+'ay'
+  return string.substring(number-1,string.length)+string.substring(0,number)+'ay'
 }
--- a/assessments/unit-01-booleans-and-strings/code.js
+++ b/assessments/unit-01-booleans-and-strings/code.js
@@ -29,3 +29,7 @@ const firstAndLast = (string) => {
 const swapFrontAndBack = (string) => {
   return string.substring(string.length/2, string.length) + string.substring(0,string.length/2)
 }
+
+const simplePigLatin = (string,number) => {
+  return string.substring(number-1,string.length)+string.substring(0,number-1)+'ay'
+}
--- a/assessments/unit-01-booleans-and-strings/code.js
+++ b/assessments/unit-01-booleans-and-strings/code.js
@@ -31,5 +31,5 @@ const swapFrontAndBack = (string) => {
 }
 
 const simplePigLatin = (string,number) => {
-  return string.substring(number-1,string.length)+string.substring(0,number)+'ay'
+  return string.substring(number,string.length)+string.substring(0,number-1)+'ay'
 }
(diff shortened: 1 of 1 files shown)


assessments/unit-01-booleans-and-strings/code.js
```diff
@@ -9,3 +9,31 @@ const rand = (n) => Math.floor(Math.random() * n);
 // syntactically correct function skeleton, more credit for a skeleton with a
 // reasonable argument list, yet more credit for a working function, and full
 // credit for a clear and simple working function.
+
+const fireAlarm = (pulled,smoke,planned) => {
+  return !pulled || !smoke || !planned
```
gigamonkey, line 14: I'm not sure why you put a `!` in front of all of these. This function would have alarm *always* go off unless all three conditions happed to be true. If you take out all the `!`s this would be correct.

assessments/unit-01-booleans-and-strings/code.js
```diff
@@ -9,3 +9,31 @@ const rand = (n) => Math.floor(Math.random() * n);
...
+}
+
+const canBePresident = (age,citizen,years) => {
+  if (age >= 35, citizen = true, years >= 14) {
```
gigamonkey, line 18: You should never compare boolean values to literals like `true` because you risk falling into the trap that you did here of using `=` (i.e. assignment) rather than `===` (comparison). Also, you need to combine the three expressions with `&&` not commas. So this should be:

```javascript
age >= 35 && citizen && years >= 14
```

Finally, whenever you have code of the form:

```javascript
if (condition) {
  return true;
} else {
  return false;
}
```

you can replace it with just:

```javascript
return condition;
```

The value in the condition part of the `if` is already a boolean so just return it.

assessments/unit-01-booleans-and-strings/code.js
```diff
@@ -9,3 +9,31 @@ const rand = (n) => Math.floor(Math.random() * n);
...
+    return false
+  }
+}
+
+const firstAndLast = (string) => {
+  return string[0]+string[string.length - 1]
+}
+
+const swapFrontAndBack = (string) => {
+  return string.substring(string.length/2, string.length) + string.substring(0,string.length/2)
```
gigamonkey, line 30: This is correct though you could leave the second argument out of the first call to `substring` because when called with just one argument the second argument defaults to the length of the string.

assessments/unit-01-booleans-and-strings/code.js
```diff
@@ -9,3 +9,31 @@ const rand = (n) => Math.floor(Math.random() * n);
...
+}
+
+const simplePigLatin = (string,number) => {
+  return string.substring(number,string.length)+string.substring(0,number-1)+'ay'
```
gigamonkey, line 34: Almost right but because the second argument to substring in exclusive (meaning the character at that index is not included) you don't need the `- 1` on the second argument to the second call to substring.

```suggestion
  return string.substring(number)+string.substring(0,number)+'ay'
```

assessments/unit-01-booleans-and-strings/code.js
```diff
@@ -9,3 +9,31 @@ const rand = (n) => Math.floor(Math.random() * n);
...
+}
+
+const randomCharacter = (string) => {
+  return string[rand]
```
gigamonkey, line 38: `rand` is a function so you need to call it meaning there need to be `()`s and, in this case, a single argument.

```suggestion
  return string[rand(string.length)]
```

gigamonkey (approved):
Better job on the string functions you tried than the booleans. You should go ahead and merge this and then you can request a new review on any more work you do on this. And I'd encourage you to try to finish the problems you didn't get to.

PR #10, opened 2022-10-31
Assessments/2022 10 31 strings

elijahsur:


assessments/2022-10-31-strings/code.js
```diff
@@ -10,3 +10,52 @@ const rand = (n) => Math.floor(Math.random() * n);
...
 // function but you can get partial credit for a function that is basically
 // correct even if contains small mistakes.
+const firstHalf = (string) => {
+  return string.substring(0, string.length / 2)
+}
+const secondHalf = (string) => {
+  return string.substring(string.length / 2, string.length)
+}
+const upDown = (string) => {
+  return string.toUpperCase() + string
```
gigamonkey, line 20: Technically needs a `toLowerCase()` on the second `string`.

assessments/2022-10-31-strings/code.js
```diff
@@ -10,3 +10,52 @@ const rand = (n) => Math.floor(Math.random() * n);
...
+}
+const firstFewEveryOther = (string) => {
+  let fin = ''
+  for (let i = 0; i > string.length - 1; i + 2) {
```
gigamonkey, line 24: Almost a good use of a loop except for the condition is swapped. Should be `<` rather than `>`. Also, the function spec only asked for the first, third, and fifth—this is more general (or would be with the fix to the loop) than what I actually asked for. (I asked that way because when I originally asked this question we hadn't covered loops yet.)

assessments/2022-10-31-strings/code.js
```diff
@@ -10,3 +10,52 @@ const rand = (n) => Math.floor(Math.random() * n);
...
+    fin + string.substring(i, i)
+  }
+  return fin
+}
+const upDownLastCharacter = (string) => {
+  return string.substring(string.length - 1, string.length).toUpperCase() + string.substring(string.length - 1, string.length)
```
gigamonkey, line 30: Needs `toLowerCase()`

gigamonkey: Also, this would be easier to read if you used `string[string.length - 1]` to get the last character rather than using `substring`.

assessments/2022-10-31-strings/code.js
```diff
@@ -10,3 +10,52 @@ const rand = (n) => Math.floor(Math.random() * n);
...
+  return string.substring(0, 1) + string.substring(string.length - 1, string.length)
+}
+const swapFrontAndBack = (string) => {
+  return string.substring(string.length / 2, string.length) + string.substring(0, string.length / 2)
+}
+const simplePigLatin = (string, number) => {
+  return string.substring(number, string.length) + string.substring(0, number) + 'ay'
+}
+const randomCharacter = (string) => {
+  return string[rand]
```
gigamonkey, line 42: So close. `rand` is a function so you need to have an argument in `()`s. In this case it should be `rand(s.length)` to get a random number between `0` and `s.length - 1`, inclusive.

assessments/2022-10-31-strings/code.js
```diff
@@ -10,3 +10,52 @@ const rand = (n) => Math.floor(Math.random() * n);
...
+}
+const randomCharacterUpDown = (string) => {
+
+}
+const isAllUpperCase = (string) => {
+  return string === string.toUpperCase()
+}
+const sameIgnoringCase = (string1, string2) => {
+  return string1 === string2
```
gigamonkey, line 51: This doesn't ignore case. One solution is `string1.toUpperCase() === string2.toUpperCase()`. I.e. smash them into all one case before comparing.

assessments/2022-10-31-strings/code.js
```diff
@@ -10,3 +10,52 @@ const rand = (n) => Math.floor(Math.random() * n);
...
+}
+const firstName = (string) => {
+  return string.substring(0, string.search(' '))
```
gigamonkey, line 54: I marked this correct because it works though I expected you to use `indexOf` rather than `search`. They both work, `search` is a bit more general.

assessments/2022-10-31-strings/code.js
```diff
@@ -10,3 +10,52 @@ const rand = (n) => Math.floor(Math.random() * n);
...
+}
+const lastName = (string) => {
+  return string.substring(string.search(' ') + 1, string.length)
+}
+const initials = (string) => {
+  return string.substring(0, 0) + string.substring(string.search(' ') + 1, string.search(' ') + 1)
```
gigamonkey, line 60: `s.substring(n, n)` (i.e. with the same value for the start and end is always an empty string. Thus this just returns the concatenation of two empty strings rather than two single letters.

gigamonkey (approved):
A few small mistakes that I commented on but overall a good job and one of the stronger performances in the class. Nice work. You can go ahead and merge this.

PR #11, opened 2022-11-08
Assignments/object functions

elijahsur:


assignments/object-functions/code.js
```diff
@@ -0,0 +1,43 @@
+const getX = (object) => {
+  return object['x']
```
gigamonkey, line 2: Yup. That works. You can also write `object.x`

assignments/object-functions/code.js
```diff
@@ -0,0 +1,43 @@
...
+}
+const point = (x, y) => {
+  let object = { "x": x, "y": y }
+  return object
```
gigamonkey, line 6: This works but you don't actually need the variable. You can just write: `return { "x": x, "y": y }`

gigamonkey (approved):
Looks good. I only commented one place but note that you don't need to assign a value to a variable in order to be able to`return` it. Also you should make sure you're comfortable with the `object.x` syntax as well as `object['x]'`. But you can go ahead and merge this.

PR #12, opened 2022-11-14
Assessments/2022 11 14 objects arrays and control constructs

elijahsur:
didn't fully grasp how to do the sum one, though i understood the prompt :(

### DIFF
--- a/assessments/2022-11-14-objects-arrays-and-control-constructs/code.js
+++ b/assessments/2022-11-14-objects-arrays-and-control-constructs/code.js
@@ -75,17 +75,17 @@ const sums = (n) => {
 
 const rule110 = (cells) => {
   let newarray = []
-  newarray[0] = 0
+  
   for (let i = 1; i < cells.length; i++) {
-    if (cells[i - 1] = 0) {
-      if (cells[i] = 0) {
-        if (cells[i + 1] = 0) {
+    if (cells[i - 1] === 0) {
+      if (cells[i] === 0) {
+        if (cells[i + 1] === 0) {
           newarray.push(0)
         }
       }
     } else {
-      if (cells[i] = 0) {
-        if (cells[i + 1] = 0) {
+      if (cells[i] === 0) {
+        if (cells[i + 1] === 0) {
           newarray.push(0)
         }
       } else {
--- a/assessments/2022-11-14-objects-arrays-and-control-constructs/code.js
+++ b/assessments/2022-11-14-objects-arrays-and-control-constructs/code.js
@@ -56,8 +56,8 @@ const bigWinners = (players) => {
 const fillTimesTable = (table) => {
   for (let i = 0; i < table.length; i++) {
     table[i][0] = (i + 1)**2
-    table[i][1] = (i + 2)**2
-    table[i][2] = (i + 3)**2
+    table[i][1] = (i + 1)**2
+    table[i][2] = (i + 1)**2
   }
   return table
 };
--- a/assessments/2022-11-14-objects-arrays-and-control-constructs/code.js
+++ b/assessments/2022-11-14-objects-arrays-and-control-constructs/code.js
@@ -6,7 +6,7 @@ const area = (rect) => {
 };
 
 const higherPaid = (e1, e2) => {
-  if (e1.salary > e2){
+  if (e1.salary > e2.salary){
     return e1
   } else {
     return e2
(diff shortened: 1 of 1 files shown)


assessments/2022-11-14-objects-arrays-and-control-constructs/code.js
```diff
@@ -2,24 +2,54 @@
 // should save you some time.
 
 const area = (rect) => {
+  return rect.width * rect.height
```
gigamonkey, line 5: Correct.

assessments/2022-11-14-objects-arrays-and-control-constructs/code.js
```diff
@@ -2,24 +2,54 @@
...
 };
 
 const higherPaid = (e1, e2) => {
+  if (e1.salary > e2.salary){
```
gigamonkey, line 9: Correct.

assessments/2022-11-14-objects-arrays-and-control-constructs/code.js
```diff
@@ -2,24 +2,54 @@
...
+  } else {
+    return e2
+  }
 };
 
 const isSamePoint = (p1, p2) => {
+  if (p1.x = p2.x) {
+    if (p1.y = p2.y) {
+      return true
+    }
```
gigamonkey, line 20: Technically you need an `else` branch here to return `false`, otherwise this function will return `undefined` in this case. Or you can avoid all this mess and write a single boolean expression that you just return:

```javascript
return p1.x === p2.x && p1.y === p2.y
```

assessments/2022-11-14-objects-arrays-and-control-constructs/code.js
```diff
@@ -2,24 +2,54 @@
...
+  } else {
+    return false
+  }
 };
 
 const totalWithTip = (bill, tipPercentage) => {
+  return {subtotal: bill.subtotal, tip: bill.subtotal*tipPercentage, total: bill.subtotal + bill.subtotal*tipPercentage}
```
gigamonkey, line 27: Correct.

assessments/2022-11-14-objects-arrays-and-control-constructs/code.js
```diff
@@ -2,24 +2,54 @@
...
+    return true
+  } else {
+    return false
+  }
 };
 
 const updateWins = (players) => {
+  for (i = 0; i < players.length; i++) {
+    if (players[i].score > 100) {
+      players[i].wins = players[i].wins + 1
```
gigamonkey, line 41: This is fine. You can also use `++` as in `players[i].wins++`.

gigamonkey (changes requested):
Looks like you're workin on this more which is great. Keep on going!

assessments/2022-11-14-objects-arrays-and-control-constructs/code.js
```diff
@@ -2,30 +2,68 @@
...
+      bigWins.push(players[i])
+    }
+  }
+  return bigWins
 };
 
 const fillTimesTable = (table) => {
 };
 
 const sums = (n) => {
```
gigamonkey, line 59: Other than the indentation, this is correct.

gigamonkey (changes requested):
`sums` looks good! Keep going. (Do `fillTimesTable` first though. `rule110` is very challenging.

assessments/2022-11-14-objects-arrays-and-control-constructs/code.js
```diff
@@ -2,31 +2,98 @@
...
+      if (cells[i] === 0) {
+        if (cells[i + 1] === 0) {
+          newarray.push(0)
+        }
+      } else {
+        newarray.push(1)
+      }
+    }
+  }
+  newarray[cells.length] = 0
```
gigamonkey, line 96: This is close—you have a lot of the right ideas but it doesn't quite fit together, I don't think. For one thing you loop to the normal end of the array, i.e. `i < cells.length` but then you try to index into `cells[i + 1]` which will be out of of bounds. And by starting the loop at `i = 1` you avoid the same problem with `cells[i - 1]` but it also means you don't make an entry in `newarray` corresponding to `cells[0]`.

But this was probably way too hard a problem for an assessment so overall a good attempt.

gigamonkey (approved):
You can go ahead and merge this.

PR #14, opened 2022-12-01
Assessments/unit 01 booleans and strings

elijahsur:


assessments/unit-01-booleans-and-strings/code.js
```diff
@@ -10,30 +10,73 @@ const rand = (n) => Math.floor(Math.random() * n);
...
-  if (age >= 35, citizen = true, years >= 14) {
-    return true
+const canBePresident = (age, citizen, years) => {
+  if (age >= 35) {
+    if (citizen === true) {
+      if (years >= 14) {
+        return true
+      } else { return false }
+    } else { return false }
+  } else { return false }
```
gigamonkey, line 24: To your credit you rembered all the `else { return false }` cases so these are correct. But much simpler would be to write a single boolean expression:

```javascript
return age >= 35 && citizen && years >= 14
```

Similar comment about several of the functions below.

assessments/unit-01-booleans-and-strings/code.js
```diff
@@ -10,30 +10,73 @@ const rand = (n) => Math.floor(Math.random() * n);
...
+  return string.substring(number, string.length) + firsthalf + 'ay'
 }
 
 const randomCharacter = (string) => {
-  return string[rand]
+  return string[rand(string.length)]
+}
+
+const randomCharacterUpDown = (string) => {
+  let random = string[rand(string.length)]
```
gigamonkey, line 74: Fine. But even slightly more stylish, use the function you just defined:

```javascript
let random = randomCharacter(string)
```

gigamonkey (approved):
You probably posted this mostly for Github purposes but left a couple quick comments anyway. You can merge this.

PR #15, opened 2022-12-02
Assessments/2022 12 02 booleans

elijahsur:


assessments/2022-12-02-booleans/code.js
```diff
@@ -0,0 +1,37 @@
+const needHeavyCoat = (goingOutside, isCold) => {
+  if (goingOutside === true) {
+    if (isCold === true) {
+      return true
+    } else { return false }
+  } else { return false }
+}
```
gigamonkey, line 7: This is correct but way more complicated than it needs to be. You should watch this video I made last year https://www.youtube.com/watch?v=yGX-OGhyf4Q which I put in classroom a few weeks ago for some ideas about how to simplify code like this.

assessments/2022-12-02-booleans/code.js
```diff
@@ -0,0 +1,37 @@
...
+const needSunscreen = (goingToBeach, goingSkiing) => {
+  return goingToBeach || goingSkiing
```
gigamonkey, line 9: This is the right way to write things like this!

assessments/2022-12-02-booleans/code.js
```diff
@@ -0,0 +1,37 @@
...
+    } else { return false }
+  } else { return false }
+}
+const isVenemous = (striped, blueHead) => {
+  if (blueHead === true) {
+    if (striped === false) {
+      return false
+    } else { return true }
+  } else { return true }
+}
```
gigamonkey, line 24: Again, this code is correct but can be much simpler.

assessments/2022-12-02-booleans/code.js
```diff
@@ -0,0 +1,37 @@
...
+const twiceAsExpensive = (price1, price2) => {
+  return price1 > price2 * 2
+}
+const winningRecord = (numberOfWins, numberOfLosses, numberOfTies) => {
+  return numberOfWins > (numberOfWins + numberOfLosses + numberOfTies) / 2
+}
+const isMagicNumber = (number) => {
+  return number === 42 || number === 17
+}
+// of course 42 is a magic number, its the answer to life, the universe, everything!
```
gigamonkey, line 37: 5,000,000 extra credit points for you!

gigamonkey (approved):
Nice job; 100% correct! But do watch that video and think about how you can write some of these in a much simpler way.

PR #17, opened 2022-12-07
Assessments/2022 12 07 control constructs

elijahsur:


assessments/2022-12-07-control-constructs/code.js
```diff
@@ -78,3 +78,54 @@ const pair = (a, b) => {
 
 ////////////////////////////////////////////////////////////////////////
 // Write your code here ...
+const logIfOk = (arg) => {
+ if (isOk(arg) === true) {
```
gigamonkey, line 82: I feel like I've pointed out before but you don't need to compare boolean values (such as the one returned by `isOk` to a literal boolean. `isOk(arg)` is going to return either `true` or `false` so comparing it to `true` is redundant. `if (isOk(arg)) ...` is sufficient.

assessments/2022-12-07-control-constructs/code.js
```diff
@@ -78,3 +78,54 @@ const pair = (a, b) => {
...
+}
+const firstOk = (arg) => {
+  for (let i = 0; i < arg; i++) {
+    if (isOk(i) === true) {
+      return i
+    }
+  }
+}
+const timeToLeet = () => {
+  for (let i = 0; i < i + 2; i++) {
```
gigamonkey, line 94: This is a clever way to create an infinite loop (that you break out of by returning.) I'd have used a `while` loop but this gets the job done. Though for maximum style `i < i + 1` feels more elegant than `i < i + 2`.

assessments/2022-12-07-control-constructs/code.js
```diff
@@ -78,3 +78,54 @@ const pair = (a, b) => {
...
+    if (isLeet(random10k()) === true) {
+      return i
+    }
+  } 
+}
+const classify = (arg) => {
+  if (isOk(arg) === true) {
+    recordOk(arg)
+  } else {recordNotOk(arg)}
```
gigamonkey, line 103: This formatting is a bit wonky but everything is correct.

assessments/2022-12-07-control-constructs/code.js
```diff
@@ -78,3 +78,54 @@ const pair = (a, b) => {
...
+  return total
+}
+const pairs = (arg) => {
+  for (let i = 1; i + 1< arg; i++) {
+    let a = i + 1
+    pair(i, i)
+    pair(i, a)
+    pair(a, i)
+    pair(a, a)
+  }
```
gigamonkey, line 130: This is the only one you missed. I think you  missed that you need to use a nested loop. Check the posted answers if you're not sure how it should work.

gigamonkey (approved):
Excellent work! Just one problem that gave you any trouble at all.

PR #18, opened 2022-12-12
Assessments/2022 12 12 arrays and objects

elijahsur:


assessments/2022-12-12-arrays-and-objects/code.js
```diff
@@ -0,0 +1,31 @@
...
+  ]
+}
+const makeMove = (mark, row, column) => {
+  return {mark: mark, row: row, column: column}
+}
+const placeMark = (array, object) => {
+  array[object.row][object.column] = object.mark
+}
+const allTheSame = (array) => {
+  return array[0][0] === array[0][1] === array[0][2]
```
gigamonkey, line 18: Two issues here. One is you can't chain the `===` operator like this, since the first `===` will evaluate to a boolean which will then be compared with `===` to the last value. You need to `&&` together separate expressions.

Also, you maybe got misled by the other functions dealing with 2d arrays but this one was intended to work on the direct elements of the array so this should be `array[0]`, `array[1]`, and `array[2]`. Putting it together

```javascript
array[0] === array[1] && array[1] === array[2]
```

would do the trick.

gigamonkey (approved):
I'm curious if you didn't get to the last problems because you couldn't make sense of them or because you spent all your time on the ones you did? Anyway, check out the posted answers and we'll go over these in class. You can go head and merge this.

PR #19, opened 2022-12-16
Assessments/2022 12 16 fall semester final

elijahsur:


gigamonkey (approved):
I'm not leaving detailed comments on individual PRs for the final but I've put all the answers students gave here with comments about which ones were correct, half correct, or incorrect:

https://github.com/bhs-intro-to-programming/answers/blob/main/assessments/2022-12-16-fall-semester-final/all-code.js

You got 28 correct and 1 half correct of the 31 questions you answered for 92% overall. Very nice work!

Here are the questions you missed or got only half right:

- willSeeTweet
- timeToLeet
- threewayClassify

You can go ahead and merge this.

PR #32, opened 2023-01-10
Projects/first web

elijahsur:


gigamonkey (changes requested):
Whoops. All your files are down a level under assignments where projects should be in the root folder. I'm not actually sure how that happened. Is `start-project` is `assignments`?

gigamonkey (approved):

PR #33, opened 2023-01-11
started work on caesar cipher

elijahsur:
wasnt sure if i should do a pull request yet, better to be safe than sorry though

### DIFF
--- a/projects/caesar-cipher/cipher.js
+++ b/projects/caesar-cipher/cipher.js
@@ -1,0 +1,13 @@
+
+let encrypt = (text, key) => {
+    let encrypted = ''
+    let key2 = 'abcdefghijklmnopqrstuvwxyz'
+    for (let i = 0; i < text.length; i++) {
+        for (let j = 0; j < key2.length; j++) {
+            if (text[i] === key2[j]) {
+                encrypted = encrypted + key2[j + key]
+            }
+        }
+    }
+    console.log(encrypted)
+}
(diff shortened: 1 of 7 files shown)


projects/caesar-cipher/cipher.js
```diff
@@ -0,0 +1,13 @@
...
+    let encrypted = ''
+    let key2 = 'abcdefghijklmnopqrstuvwxyz'
+    for (let i = 0; i < text.length; i++) {
+        for (let j = 0; j < key2.length; j++) {
+            if (text[i] === key2[j]) {
+                encrypted = encrypted + key2[j + key]
+            }
+        }
+    }
+    console.log(encrypted)
```
gigamonkey, line 12: A good way to structure and test things like this is to write a function like `encrypt` that returns a value and then _outside_ the function have a call (or several) to `console.log(encrypt('thing to encrypt'))`.

The to be a bit more fancy you can check things like:

```javascript
console.log(encrypt('foo', 0) === 'foo')) // i.e. with no shift encrypted should be the same as plaintext
console.log(encrypt(encrypt('foo', 13), 13) === 'foo')) // encrypting twice with 13 should bring you back to orig
```

And then you could get even fancier and write a function to test things and print out more useful information about what is being tested than just `true` or `false`.

gigamonkey (approved):
Good start. A couple comments below about testing things.

PR #34, opened 2023-01-12
finished encryption; decryption code not done

elijahsur:


projects/caesar-cipher/cipher.js
```diff
@@ -5,9 +5,31 @@ let encrypt = (text, key) => {
     for (let i = 0; i < text.length; i++) {
         for (let j = 0; j < key2.length; j++) {
             if (text[i] === key2[j]) {
```
gigamonkey, line 7: This loop is basically a reimlementation of the `indexOf` method on strings. I.e. you could write `key2.indexOf(text[i])` to find out where in `key2` the character at `text[i]` occurs. You have to deal with it returning -1 if it's not in `key2` but that's not too terrible.

projects/caesar-cipher/cipher.js
```diff
@@ -5,9 +5,31 @@ let encrypt = (text, key) => {
...
     }
-    console.log(encrypted)
-}
\ No newline at end of file
+    return encrypted
+}
+
+let decrypt = (text, key) => {
+    let decrypted = ''
+    let key2 = 'abcdefghijklmnopqrstuvwxyz'
```
gigamonkey, line 23: Since both `decrypt` and `encrypt` need this value, it make make sense to declare it outside both functions as a global variable. And maybe name it a bit more descriptively: `alphabet` maybe. At any rate `key` in the context of encryption would usually refer to the secret key you use to do the encryption/decryption.

projects/caesar-cipher/cipher.js
```diff
@@ -5,9 +5,31 @@ let encrypt = (text, key) => {
     for (let i = 0; i < text.length; i++) {
         for (let j = 0; j < key2.length; j++) {
             if (text[i] === key2[j]) {
-                encrypted = encrypted + key2[j + key]
+                let change = j
+                if (change + key > 26) {
+                    change = 27 - j
+                } else {
+                    change = j + key
+                }
```
gigamonkey, line 13: You might want to think back to the `%` operator and what it does.

projects/caesar-cipher/cipher.js
```diff
@@ -5,9 +5,31 @@ let encrypt = (text, key) => {
...
             }
         }
     }
-    console.log(encrypted)
-}
\ No newline at end of file
+    return encrypted
+}
+
+let decrypt = (text, key) => {
```
gigamonkey, line 21: You may want to think about what the difference between encryption and decryption is in a symmetric cipher like the Caesar cipher.

gigamonkey (approved):
Good start! Left some comments of some stuff for you to think about.

PR #35, opened 2023-01-13
finished decryption; caesar cipher done

elijahsur:


elijahsur:
nevermind, i found a problem with my code

current code works with a few keys, but not all

projects/caesar-cipher/cipher.js
```diff
@@ -1,17 +1,17 @@
+let alphabet = 'abcdefghijklmnopqrstuvwxyz'
 
 let encrypt = (text, key) => {
     let encrypted = ''
-    let key2 = 'abcdefghijklmnopqrstuvwxyz'
     for (let i = 0; i < text.length; i++) {
-        for (let j = 0; j < key2.length; j++) {
-            if (text[i] === key2[j]) {
+        for (let j = 0; j < alphabet.length; j++) {
```
gigamonkey, line 6: (Hmmm. I could have sworn I left some comments yesterday but I don't see them on this PR. Maybe I never finished submitting my review.) Anyway, this loop is doing some work very similar to what the `indexOf` method on strings does: finding the position in `alphabet` where `text[i]` occurs. Can you use that method to find the position and get rid of the loop.

projects/caesar-cipher/cipher.js
```diff
@@ -20,16 +20,25 @@ let encrypt = (text, key) => {
 
 let decrypt = (text, key) => {
```
gigamonkey, line 21: Decrypting should just beencrypting with the original key * -1. Can you write `encrypt` so it handles negative keys and then write `decrypt` as a one-line function that just calls `encrypt` with the key negated?

gigamonkey (approved):
Good progress. Though some comments for you to think about.

gigamonkey:
Oh, I see. This is a new PR which is why I didn't see my old comments. Which is good. I just got fooled by the similar PR titles.

PR #36, opened 2023-01-17
text done on DOM assignment, need help with rest

elijahsur:


gigamonkey (approved):
Good start. As I said I'll talk about the issue you got stuck on. You can merge this at the beginning of class but don't delete the branch. Keep working on that branch in your Codespace and then you can Commit and Push today's work and you should be able to make a new PR of just today's charges. Let me know, of course, if that doesn't seem to be working.

PR #37, opened 2023-01-18
finished list; will work on image tomorrow

elijahsur:


gigamonkey (approved):
Looks good. We'll talk later in class about how to make code like this less cumbersome by writing a few functions. But you've clearly got the basic idea of how to do the DOM manipulations. You can merge this and then make a new PR for the rest.

PR #38, opened 2023-01-19
all except formatting of credits;dont know how

elijahsur:


gigamonkey (approved):
This looks good. I can't remember if I suggested to you to look at the source HTML of the static version which is in `public/index.html`. If you do, you should be able to see the structure of the credits: the two `p` elements are inside a `div` element that has a `class` attribute set. If you recreate that structure, then the css attached to this file will style the credits for you.

But you can go ahead and  merge this but don't delete the branch as I may have you work on it just a bit more.

PR #39, opened 2023-01-20
credits done; think everything is done

elijahsur:


projects/2023-01-17/public/scripts/script.js
```diff
@@ -1,14 +1,35 @@
...
+  return e;
+};
+
+const list = (codetext, text) => {
+  const e = document.createElement('li');
+  const code = document.createElement('code');
+  code.append(codetext)
+  e.append(code)
+  e.append(
+    document.createTextNode(`: These elements are used for headings and subheadings. <h1> is the highest level heading, while <h6> is the lowest.
```
gigamonkey, line 13: I assume you were on the way to replacing this literal string with the `text` variable.

projects/2023-01-17/public/scripts/script.js
```diff
@@ -1,14 +1,35 @@
+const cel = (tag, text) => {
```
gigamonkey, line 1: Now that you've got this function, you should find all the places you can use it and do so. Both because that's how you get the benefit of the function but also it may reveal either changes you want to make to this function or possibly other functions you want to write (since some cases won't be covered by this one.)

projects/2023-01-17/public/scripts/script.js
```diff
@@ -1,14 +1,35 @@
...
+  const e = document.createElement(tag);
+  e.append(document.createTextNode(text));
+  return e;
+};
+
+const list = (codetext, text) => {
```
gigamonkey, line 7: This is another good function. Though definitely take my advice about the `cel` function and as soon as you've got it working try to use it wherever you can. I think you'll find it useful in most cases and will also find a few cases where it doesn't quite do what you need. Then the name of the game is deciding what to do about those cases. (There's no one right answer.)

gigamonkey (approved):
Good start on some helper functions. You can merge this but don't delete the branch as I may ask you to work on it some more. (I haven't totally decided yet though I may have decided by the time you read this.)

PR #40, opened 2023-01-23
list partly less crowded, used functions

elijahsur:


projects/2023-01-17/public/scripts/script.js
```diff
@@ -10,164 +10,53 @@ const list = (codetext, text) => {
...
-li10.append(code10)
-li10.append(
-  document.createTextNode(`: These elements are used for grouping and applying styles to elements on a web page.
-  `)
-)
-a3.append(li1,li2,li3,li4,li5,li6,li7,li8,li9,li10);
+a3.append(list('<html>',`: This is the root element of an HTML document, and it contains all of the other elements.`))
+a3.append(list('<head>',`: This element contains information about the document, such as the title of the page, which is displayed in the browser's title bar or tab.`))
+a3.append(list(`<body>`,`: This element contains the content of the document that is displayed in the browser window.`))
+a3.append(list(`<h1> to <h6>`,`: These elements are used for headings and subheadings. <h1> is the highest level heading, while <h6> is the lowest.`))
```
gigamonkey, line 52: Technically this one is not 100% correct since only the tags `<h1>` and `<h6>` should be in `<code>` while the " to " should not be. But the `list` function is a good one.

projects/2023-01-17/public/scripts/script.js
```diff
@@ -10,164 +10,53 @@ const list = (codetext, text) => {
...
+a3.append(list(`<p>`,`This element is used for paragraphs of text.`))
+a3.append(list(`<a>`,`This element is used for hyperlinks, which allow users to navigate between pages on the web.`))
+a3.append(list(`<img>`,`: This element is used to embed images in a web page.`))
+a3.append(list(`<ul> and <ol>`,`: These elements are used for unordered and ordered lists, respectively.`))
```
gigamonkey, line 56: Same problem as the previously commented line.

gigamonkey (approved):
Nice work. You can merge this and start looking at the web-demos.

PR #41, opened 2023-01-24
Projects/web demos

elijahsur:


projects/web-demos/public/click-counter/script.js
```diff
@@ -1,13 +1,24 @@
...
+    .fill()
+    .map(() => Math.floor(Math.random() * 255));
+//i want to try to add a random color to the number every 10 numbers, but im not sure how to use this randomRGB function
+const rgb = (r, g, b) => `rgb(${r}, ${g}, ${b})`;
 document.querySelector('p').onclick = (e) => {
   const p = e.target;
   counter++;
   if (counter === 1) {
-    p.innerHTML = 'I have been clicked <span id="counter"></span> time<span id="s"></span>.';
+    p.innerHTML = 'I have been clicked <span id="counter"></span> time</span>.';
```
gigamonkey, line 12: You left behind an extra closing `</span>` tag.

projects/web-demos/public/click-counter/script.js
```diff
@@ -1,13 +1,24 @@
...
   }
-  if (counter === 2) {
-    p.querySelector('#s').textContent = 's';
+  if (counter > 1) {
+    p.innerHTML = 'I have been clicked <span id="counter"></span> times</span>.'
+  }
+  if (counter/10 === Math.floor(counter/10)) {
```
gigamonkey, line 17: I would probably write that:

```javascript
if (counter % 10 === 0) {
```

as using the `%` operator is the idiomatic way to check whether a number is divisible by another number.

projects/web-demos/public/click-counter/script.js
```diff
@@ -1,13 +1,24 @@
...
+    p.innerHTML = 'I have been clicked <span style="color:red"><span id="counter"></span></span> times</span>!'
+  }
+  if (counter/100 === Math.floor(counter/100)) {
```
gigamonkey, line 20: Likewise but `% 100`

projects/web-demos/public/click-counter/script.js
```diff
@@ -1,13 +1,24 @@
 let counter = 0;
-
+const randomRGB = () =>
+  Array(3)
+    .fill()
+    .map(() => Math.floor(Math.random() * 255));
+//i want to try to add a random color to the number every 10 numbers, but im not sure how to use this randomRGB function
```
gigamonkey, line 6: So `randomRGB` returns an array of three values representing the red, green, and blue amounts. (Sorry, I wrote it in a bit of an advanced style so that may not be obvious.) `rgb` is a function that takes three such values but as separate arguments. You could put them together like this knowing what you know now:

```javascript
const randomRGBColor = () => {
  const values = randomRGB();
  return rgb(values[0], values[1], values[2]);
};
``` 

Then you'll need to interpolate that string in place of the `red` and `yellow` you used below.

gigamonkey (changes requested):
This is a nice little attempt at extending the existing code. Exactly the kind of thing I was looking for. Why don't you keep going on this and see if you can take the hint I gave you and figure out how to do the color changing every 10 clicks.

projects/web-demos/public/click-counter/script.js
```diff
@@ -1,24 +1,43 @@
...
+    p.innerHTML = `I have been clicked <span style="color:${randomRGBColor()}"><span id="counter"></span></span> times!`
+    if (counter === 10) {
+      lastClicked.textContent = 'I last reached 10 0 seconds ago!'
+      seconds = date.now() / 1000
+    } else {
+    let currentSeconds = Date.now() / 1000
+    let difference = currentSeconds - seconds
+    seconds = Date.now() / 1000
+    lastClicked.textContent = 'I last reached 10 ' + difference + ` seconds ago!`
+    // how would i make "difference" only include the decimal spots up to the first 2-ish? just to simplify the number
```
gigamonkey, line 36: There's a method on numbers `toFixed` that returns a string representing the number to a given number of decimal places:

```javascript
const t = Date.now();
// wait a bit
console.log(((Date.now() - t)/1000).toFixed(1))
```

gigamonkey (approved):
Nice work coming up with little changes to make. Exactly the goal here. Feel free to merge this or just keep working here and request another review when you've got some more stuff.

PR #42, opened 2023-01-26
Projects/genetic algorithm

elijahsur:


gigamonkey (approved):
This is a good start—you have at least one key part done which is generating a completely random string. I'm not quite sure what you're going for with lines 13-34 but given that we haven't had a chance to talk about how genetic algorithms work that's maybe understandable. Let's do that on Monday.

Also, you probably should move `algorithms.js` into the `public/scripts` directory to keep things organized.

projects/genetic-algorithm/public/scripts/algorithm.js
```diff
@@ -0,0 +1,26 @@
+let alphabet = 'abcdefghijklmnopqrstuvwxyz '
+let algorithm = (text) => {
+    // generates random list of characters the same length as the text
+    let generation = ''
+    for (let i = 0; i < text.length; i++) {
+        generation = generation + alphabet[Math.floor(Math.random() * 27)]
+    }
+    let genarray = []
+    let genversion = 0
+    genarray.push(generation)
```
gigamonkey, line 10: Since arrays know how long they are, you don't really need the `genversion` variable. When you want to get the last element of `genarray` just use `genarray[genarray.length - 1]`. This will simplify your code because you don't have to worry about keeping `genversion` and the length of `genarray` in sync.

projects/genetic-algorithm/public/scripts/algorithm.js
```diff
@@ -0,0 +1,26 @@
...
+    let limit = 1000;
+    while (genarray[genversion] != text && limit-- > 0) {
+        let gen = ''
+        for (let j = 0; j < genarray[genversion].length; j++) {
+            if (genarray[genversion][j] === text[j]) {
+                gen = gen + genarray[genversion][j]
```
gigamonkey, line 16: You can use `+=` with strings to. So you can write this as:

```javascript
gen += genarray[genversion][j]
```

projects/genetic-algorithm/public/scripts/algorithm.js
```diff
@@ -0,0 +1,26 @@
...
+            } else {
+                gen = gen + alphabet[Math.floor(Math.random() * 27)]
```
gigamonkey, line 18: And `+=` here too.

projects/genetic-algorithm/public/scripts/algorithm.js
```diff
@@ -0,0 +1,26 @@
+let alphabet = 'abcdefghijklmnopqrstuvwxyz '
+let algorithm = (text) => {
+    // generates random list of characters the same length as the text
+    let generation = ''
+    for (let i = 0; i < text.length; i++) {
+        generation = generation + alphabet[Math.floor(Math.random() * 27)]
+    }
```
gigamonkey, line 7: These three lines are worth breaking out into a separate function. `makeRandomPhrase` or something. I'd probably write it to take the desired length as an argument since that all we really care about.

projects/genetic-algorithm/public/scripts/algorithm.js
```diff
@@ -0,0 +1,26 @@
...
+    let genarray = []
+    let genversion = 0
+    genarray.push(generation)
+    let limit = 1000;
+    while (genarray[genversion] != text && limit-- > 0) {
```
gigamonkey, line 12: There are a lot of occurrences of the expression `genarray[genversion]` in the next ten or so lines of code. Since initially that expression has the same value as the variable `generation`, I'd probably  use the variable everywhere it occurs and then arrange to reassign `generation` to the new value at the bottom of the loop. Basically, you use the original `generation` while you build up `gen` and then once you're done building it up, you can swap it in:

```javascript
generation = gen
```

And then push it onto `genarray` since you want to keep a collection of the different generations.

gigamonkey (approved):
Glad you got this working. I left some comments just in general about how to make the code simpler which you should try to grok but Tuesday I'll try to make some time to talk about genetic algorithms and then you'll change a bunch of this code anyway.

projects/genetic-algorithm/public/scripts/algorithm.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+    while (done === false) {
+        // will determine the top 25% of the batch
+        let temparray = []
+        for (let j = 0; j < temp.length; j++) {
+            temparray.push[temp[j][1]]
+            for (let k = 0; k < temparray.length; k++) {
+                if (temparray[j][1] > temparray[k][1]) {
+                    temparray.index(k) = temparray[j]
+                }
```
gigamonkey, line 38: All these "temps" are making my head hurt. There's gotta be a more descriptive name for these variables.

gigamonkey (approved):
More good progress. Keep going! You can merge or not as you prefer. If you do merge, don't delete the branch and then just keep working.

projects/genetic-algorithm/public/scripts/algorithm.js
```diff
@@ -0,0 +1,48 @@
...
+            }
+        }
+
+        done = true
```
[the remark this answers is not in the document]
elijahsur: just temporary, was made to test the beginning parts of the code

PR #43, opened 2023-02-01
reorganized code, working on top25 function

elijahsur:


projects/genetic-algorithm/public/scripts/algorithm.js
```diff
@@ -1,48 +1,71 @@
...
+// creates the first batch of random strings
+const firstbatch = (batchamount,text) => {
+    let batcharray = []
+    for (let i = 0; i < batchamount; i++) {
+        let phrase = randomPhrase(text.length)
+        batcharray.push([phrase, fitness(phrase, text)])
+    }
+    return batcharray
+}
+// judges the fitness of a group of strings
```
gigamonkey, line 25: The fitness function should be judging the fitness of a single string. `generated` is maybe not the best name for that since lots of things could be generated. But I'd expect this function to be called with one element of the population so maybe `phrase` or `critter` or `dna` or whatever you want to call the things.

projects/genetic-algorithm/public/scripts/algorithm.js
```diff
@@ -1,48 +1,71 @@
...
 const fitness = (generated, text) => {
     let fitness = 0
     for (let i = 0; i < text.length; i++) {
-        if (generated[i] === text[i]) {
+        if (generated[i][0] === text[i]) {
```
gigamonkey, line 29: I'm not sure what that `[0]` is doing. I'd expect the line you had before where you were comparing `generated[i]` to `text[i]`.

projects/genetic-algorithm/public/scripts/algorithm.js
```diff
@@ -1,48 +1,71 @@
...
             fitness++
         }
     }
     return fitness/text.length
 }
+// determines the top 25% of a batch, with "string" being the desired string. currently reorders the batch from highest to lowest
+const top25 = (batch) => {
+    let top25 = []
+    for (let i = 0; i < batch.length; i++) {
+        for (let j = 0; j < top25.length; j++) {
```
gigamonkey, line 39: You should think about what `top25.length` will be the first time you get to this line and what implications that has about how often this loop will execute.

projects/genetic-algorithm/public/scripts/algorithm.js
```diff
@@ -1,48 +1,71 @@
...
-    let temp = []
+    let batcharray = []
     let finalarray = []
     let done = false
     //provides the first batch
-    for (let i = 0; i < batch; i++) {
-        let phrase = randomPhrase(text.length)
-        temp.push([phrase, fitness(phrase, text)])
-    }
+    firstbatch(batch)
```
gigamonkey, line 56: `firstbatch` takes two arguments but you're only passing one.

gigamonkey (changes requested):
Looks like you took a bunch of my suggestions; that's good. ;-) Looks like you have a few bugs your tracking down. I left a few comments that should point you in the right direction on a few of them.

projects/genetic-algorithm/public/scripts/algorithm.js
```diff
@@ -1,48 +1,60 @@
...
+    }
+    return generation
 }
-const fitness = (generated, text) => {
+// creates the first batch of random strings
+const firstbatch = (batchamount, text) => {
+    let batcharray = []
+    for (let i = 0; i < batchamount; i++) {
+        let phrase = randomPhrase(text.length)
+        batcharray.push([phrase, fitness(phrase, text)])
```
gigamonkey, line 21: This would be a great place to use an object. Rather than this two-element array where you need to remember that the 0th element is the phrase and the 1st element is the fitness, you could use an object:

```javascript
        batcharray.push({ phrase: phrase, fitness: fitness(phrase, text) });
```

Then elsewhere you'll be able to say, `whatever.fitness` which will be much more understandable than `whatever[1]`. Names are better than numbers for humans.

projects/genetic-algorithm/public/scripts/algorithm.js
```diff
@@ -1,48 +1,60 @@
...
+for (let i = 0; i < batch.length; i++) {
+    fitnessOnly.push(batch[i][1])
 }
+console.log(fitnessOnly.length)
+for (let j = 0; j < Math.floor(fitnessOnly.length / .25) ; j++) {
+    fitnessOnly.sort()
+    fitnessOnly.pop(j)
+}
+console.log(fitnessOnly)
+}
```
gigamonkey, line 48: First off, please run the "Format document" command to fix this indentation.

This function isn't quite right as you're extracting the fitness values but that means you're losing the association with the item that has each fitness value which you'll need later. The trick is you need to use the version of `sort()` where you pass a function that provides the sorting comparison. To give an example if I wanted to sort a bunch of video game score objects that each have `initials` and `score` properties I might write:

```javascript
games.sort((a, b) => a.score - b.score);
```

The function `(a, b) => a.score - b.score` takes two arguments and extracts the `score` property from each and subtracts them, giving me a negative number when `a.score` is less than `b.score`, 0 when they're the same, and a positive number when `a.score` is greater than `b.score`, which is exactly the contract for a comparator function.

Hopefully the translation to your specific case is obvious.

Then the other part of this function, extracting the top 25% you have almost right: you could indeed loop and remove the extra (though you need to pop off 75% to get the top 25%). Alternatively you could loop a quarter of the way through and push only those items onto a new list. Or even easier you can use the `slice` method which is to arrays as `substring` is to strings.

```javascript
sortedGames.slice(0, Math.floor(sortedGames.length * 0.25))
```

will do the trick. It'd probably be worth it to see if you can get one of the loop versions working but then you should just use `slice`.

projects/genetic-algorithm/public/scripts/algorithm.js
```diff
@@ -1,48 +1,60 @@
...
+// i think this code will work, im just not sure how to make it sort it from highest to lowest, not lowest to highest
+// currently, it deletes the highest numbers, but keeps the lowest
+const top25 = (batch) => {
+let fitnessOnly = []
+for (let i = 0; i < batch.length; i++) {
+    fitnessOnly.push(batch[i][1])
 }
+console.log(fitnessOnly.length)
+for (let j = 0; j < Math.floor(fitnessOnly.length / .25) ; j++) {
```
gigamonkey, line 43: You probably want either `length * 0.25` or `length / 4` but not `length / 0.25`.

projects/genetic-algorithm/public/scripts/algorithm.js
```diff
@@ -1,48 +1,60 @@
...
-    let temp = []
+    let batcharray = []
     let finalarray = []
     let done = false
     //provides the first batch
-    for (let i = 0; i < batch; i++) {
-        let phrase = randomPhrase(text.length)
-        temp.push([phrase, fitness(phrase, text)])
-    }
+    firstbatch(batch, text)
```
gigamonkey, line 55: You need to do something with the value returned by `firstbatch`.

gigamonkey (approved):
Some issues but I gave you some pointers and I'm sure you can figure it out.

projects/genetic-algorithm/public/scripts/algorithm.js
```diff
@@ -1,48 +1,65 @@
...
-        let temparray = []
-        for (let j = 0; j < temp.length; j++) {
-            temparray.push[temp[j][1]]
-            for (let k = 0; k < temparray.length; k++) {
-                if (temparray[j][1] > temparray[k][1]) {
-                    temparray.index(k) = temparray[j]
-                }
+        batchHistory.push(currentBatch[currentBatch.length - 1])
+        for (let i = 0; i < batchlength; i++) {
+            let phrase = currentBatch[i].phrase.slice(0, Math.floor(Math.random() * string.length)) + currentBatch[i + 1].phrase.slice(0, Math.floor(Math.random() * string.length))
```
gigamonkey, line 48: Since `currentBatch` is by definition a quarter the length of the original `batchlength` `i` is not always going to be a legal index into `currentBatch`. You can change the loop to loop up to `currentBatch.length`  and then for each position generate four children. (You also would need to watch out for `i + 1` as an index; looks like you'd want to loop up to `currentBatch.length - 1` so `i + 1` is always a legal index.

Another possibility is to treat the top 25% as prospective parents and then just pick random pairs. In that case you can loop up to `batchlength` because that's how many new critters you want to make and you'll make one per time through the loop.

gigamonkey: Another comment on this line (and elsewhere), you have a *lot* of instances of the pattern `Math.floor(Math.random() * something)`. Which is a great pattern but worth breaking out into a function that returns a random number from 0 to n which you can then use all these places which will make things a bit easier to read and thus see what's really going on.

gigamonkey: did you mean to always take the front part of both parents? That seems not quite right. In fact, for this problem you probably want to conserve the length of the critters so I'd pick one point at random and then split both parents at that same point and put together the front of one parent and the back of the other to make the child.

projects/genetic-algorithm/public/scripts/algorithm.js
```diff
@@ -1,48 +1,65 @@
...
+            phrase = phrase.slice(0, string.length)
```
gigamonkey, line 49: I was wondering what this line was for but I see now it's probably an attempt to make the children the right length since the way you built them didn't do that. Unfortunately this will only work to chop down too-long children but you can still get too-short children.

(In genetic algorithms you need to think about what part of the solution space you want to explore—you could explore the space of all possible strings, of any length, but since you know that only those strings that are the same length as the target string can possibly be the solution, that's probably a waste. It also depends on how you are producing the children and measuring fitness—if your fitness function considered `'pixzxza'` a fairly fit critter when trying to make `'pizza'` because it actually contains the whole target string in order, just with a few extra characters, and there was some way for crossing that string with another string that would result in the extra characters being dropped, then maybe it would be okay to let strings get longer and shorter than the target length. So, there's no hard-and-fast rule but I'd say your life would be simpler, in this program, if you arranged your crossing mechanism to jus always produce offspring that are the right length.)

projects/genetic-algorithm/public/scripts/algorithm.js
```diff
@@ -1,48 +1,65 @@
...
+            if (Math.floor(Math.random() * (4 - 1) + 1) === 1) {
```
gigamonkey, line 50: I assume this is aimed at producing a 1 in 3 chance of a mutation. Since `Math.random()` returns a number between 0 and 1 the normal idiom for something occuring with probability p is:

```javascript
if (Math.random() < p) {
```

In this case you can use `Math.random() < 1/3` or `Math.random() < 0.33` (though neither is *exactly* a 1/3 chance since 0.33 is clearly an approximation and 1/3 is also as it's a repeating fraction.)

gigamonkey (approved):
Good work though there are still some bugs. Which is pretty understandable because there's a lot going on here and it's hard to keep all the details straight. The best advice I can give you is be liberal in breaking things out into functions because it both makes the code easier to read but also easier to focus on one thing at a time. For instance, if you wrote a function:

```javascript
const makeChild = (parent1, parent2) => {
   // whatever
};
```

It would probably be easier to focus on what it is doing and to ask yourself if that is actually how you want to make a child from two parents than if the same code was embedded in some complicated loop with other things going on.

Anyway, keep at it. You can merge (but keep the branch) if you want or keep working on this PR.

PR #45, opened 2023-02-07
put code into a website

elijahsur:
didnt change any of the code, just integrated it into a website

### DIFF
--- a/projects/genetic-algorithm/public/scripts/script.js
+++ b/projects/genetic-algorithm/public/scripts/script.js
@@ -1,6 +1,7 @@
 ////////////////////////////////////////////////////////////////////////////////
 // Genetic Algorithm Code
 let key = prompt('enter the possible characters the algorithm can use','abcdefghijklmnopqrstuvwxyz ABCDEFGHIJKLMNOPQRSTUVWXYZ.,!?"1234567890#')
+let userbatch = prompt('enter amount per batch', 1000000)
 
 // constructs a random phrase, with "length" as the length of the string
 const randomPhrase = (length) => {
@@ -37,9 +38,9 @@ const top25 = (batch) => {
 // combines two strings next to each other to create a child, randomly mutates
 const evolve = (currentBatch, i, string) => {
   let randomPlace = random(string.length)
-  let phrase = currentBatch[i].phrase.slice(0, randomPlace) + currentBatch[i + 1].phrase.substring(randomPlace, currentBatch[i + 1].phrase.length)
+  let phrase = currentBatch[i].phrase.substring(0, randomPlace) + currentBatch[i + 1].phrase.substring(randomPlace)
   if (Math.random() < 1 / 3) {
-    phrase.replace(phrase[random(phrase.length)], key[random(key.length)])
+    phrase = phrase.replace(phrase[random(phrase.length)], key[random(key.length)])
   }
   return phrase
 }
@@ -48,11 +49,11 @@ const random = (upto) => {
   return Math.floor(Math.random() * upto)
 }
 
-let algorithm = (string, batchlength) => {
+const algorithm = (string, batchlength) => {
   let done = false
   let batchHistory = []
   let currentBatch = top25(firstbatch(string, batchlength))
-  while (done === false) {
+  while (!done) {
     batchHistory.push(currentBatch[currentBatch.length - 1])
     for (let i = 0; i < batchlength; i++) {
       let phrase = evolve(currentBatch, i, string)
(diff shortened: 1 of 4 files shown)


projects/genetic-algorithm/public/scripts/script.js
```diff
@@ -1,26 +1,70 @@
...
+}
+// ranks the fitness of all strings, then returns the top 25% of the highest fitness
+const top25 = (batch) => {
+  batch.sort((a, b) => a.fitness - b.fitness)
+  return batch.slice(Math.floor(batch.length * (3 / 4)), batch.length)
+}
+// combines two strings next to each other to create a child, randomly mutates
+const evolve = (currentBatch, i, string) => {
+  let randomPlace = random(string.length)
+  let phrase = currentBatch[i].phrase.slice(0, randomPlace) + currentBatch[i + 1].phrase.substring(randomPlace, currentBatch[i + 1].phrase.length)
```
gigamonkey, line 40: This line shows how convenient it is that if you leave out the second argument to `substring` it defaults to the length of the string. So you can leave out that cumbersome `currentBatch[i + 1].phrase.length`:

```javascript
let phrase = currentBatch[i].phrase.slice(0, randomPlace) + currentBatch[i + 1].phrase.substring(randomPlace);
```

Also, just to be consistent you should either use `slice` in both calls or `substring` in both calls. `slice` also defaults the second argument to the length of the string if it is omitted.

projects/genetic-algorithm/public/scripts/script.js
```diff
@@ -1,26 +1,70 @@
...
+const random = (upto) => {
+  return Math.floor(Math.random() * upto)
+}
 
-// Do the same thing as before but in a loop
-for (let i = 0; i < 10; i++) {
-  const p = document.createElement('p');
-  p.append(document.createTextNode(`Counting ${i}.`));
-  document.querySelector('body').append(p);
+let algorithm = (string, batchlength) => {
```
gigamonkey, line 51: Functions should almost always be defined with `const` not `let`. Using `let` implies that the variable `algorithm` will be assigned a new value at some point.

projects/genetic-algorithm/public/scripts/script.js
```diff
@@ -1,26 +1,70 @@
...
+  let done = false
+  let batchHistory = []
+  let currentBatch = top25(firstbatch(string, batchlength))
+  while (done === false) {
```
gigamonkey, line 55:
```suggestion
  while (!done) {
```

projects/genetic-algorithm/public/scripts/script.js
```diff
@@ -1,26 +1,70 @@
...
+const top25 = (batch) => {
+  batch.sort((a, b) => a.fitness - b.fitness)
+  return batch.slice(Math.floor(batch.length * (3 / 4)), batch.length)
+}
+// combines two strings next to each other to create a child, randomly mutates
+const evolve = (currentBatch, i, string) => {
+  let randomPlace = random(string.length)
+  let phrase = currentBatch[i].phrase.slice(0, randomPlace) + currentBatch[i + 1].phrase.substring(randomPlace, currentBatch[i + 1].phrase.length)
+  if (Math.random() < 1 / 3) {
+    phrase.replace(phrase[random(phrase.length)], key[random(key.length)])
```
gigamonkey, line 42: This actually isn't doing anything so I guess you learned mutation isn't super necessary. Strings in Javascript are immutable so `replace` returns a new string. To have this take effect you need to assign that new string back to `phrase`:

```javascript
    phrase = phrase.replace(phrase[random(phrase.length)], key[random(key.length)])
```

Also, that replacement may not be doing quite what you expect since `replace` produces its new string by replacing the first occurrence of the given string it finds. So `phrase[random(phrase.length)]` picks a random character from anywhere in the string but always replaces the first occurrence of that character in the string.

To replace a character at a specific place, you can use the `splice` method, not to be confused with `slice`:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice

projects/genetic-algorithm/public/scripts/script.js
```diff
@@ -29,14 +73,15 @@ for (let i = 0; i < 10; i++) {
...
-  const p = document.createElement('p');
-
-  // Create the text of the paragrph, using the event object we were passed
-  const text = document.createTextNode(`Clicked on “${e.target.parentNode.innerText}”`);
-
-  // Add the text to the paragraph
-  p.append(text);
-
+  let userstring = prompt('enter string to generate to')
+  let userbatch = prompt('enter amount per batch', 1000000)
```
gigamonkey, line 77: A next step that would be nice would be to replace these calls  to `prompt` with something that gets the user input from elements within the page. If you haven't yet, make a new project from the `web-demos` kit and look at the ui-examples page to get some ideas about how you might do that.

gigamonkey (approved):
More great work. Left some suggestions for next steps and a few things you'll want to clean up. Also note I made a PR against this PR (turtles all the way down) that you should merge into this before you merge this one into main.

gigamonkey (changes requested):
Looks good; I assume you're still working on getting rid of those calls to `prompt`.

PR #47, opened 2023-02-09
used import export; working on button

elijahsur:


gigamonkey (approved):
As usual, very nice work. One thing you can do to make the code a lot simpler is break up that event handler. You figured out a way to make it work which is pretty impressive but in a case like this where each element wants to do a different thing it would be easier to add separate event handlers to each input element that just does the appropriate actions for that element. I think you got misled by my example code where I wrote the `addOnChange` function because I wanted each input to have essentially the same event handler. But in your case, the one event handler you add to each element immediately tries to figure out which element it's being triggered by and then does something different.

gigamonkey:
If you want another medium to large challenge, look at https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API I assume you've noticed that your output only comes out in the web page when everything is done and I'm guessing you've wondered if there's a way to have it update incrementally. The issue is that Javascript can, in general, only do one thing at a time. So when your genetic algorithm code is in its main loop nothing else can happen on the page until the loop ends. There are ways you could restructure your code to not use a loop but web workers give you a way to have another thread of control running independent of the main code in your page. You can communicate with that other code by sending it data (such as the starting parameters for the GA) and it can post messages back to your code which come in as events that you can write handlers for. So you could have the GA run in a web worker that contains the main loop but posts intermediate results back to your main code which then can update the page incrementally.

This might be a bit much to bite off today but if you get intrigued, look at that page. Some ways down the page there are links to some very simple examples that you can look at and adapt.

PR #50, opened 2023-02-21
Assignments/refactor tic tac toe

elijahsur:
ive created functions out of (i think) three parts of the original code

not sure what to break out more

### DIFF
--- a/assignments/refactor-tic-tac-toe/code.js
+++ b/assignments/refactor-tic-tac-toe/code.js
@@ -52,7 +52,6 @@ const checkWinner = (r,c,lines,board) => {
       const m2 = board[r][c];
       if (m0 !== '' && m0 === m1 && m0 === m2) {
         winner = lines[i];
-        return winner
       }
     }
 }
@@ -120,7 +119,7 @@ registerOnclick((x, y) => {
     // Check if there's a winner now
     winner = null;
     checkWinner(r,c,lines,board)
-    if (checkWinner(r,c,lines,board) !== null) {
+    if (winner !== null) {
       // Draw the line through three in a row
       const [r1, c1] = winner[0];
       const [r2, c2] = winner[winner.length - 1];
--- a/assignments/refactor-tic-tac-toe/code.js
+++ b/assignments/refactor-tic-tac-toe/code.js
@@ -50,7 +50,9 @@ const board = [
   ['', '', ''],
 ];
 
-const mark = (r,c) => {
+const mark = () => {
+  let r = Math.floor((y - boardTop) / cellSize);
+  let c = Math.floor((x - boardLeft) / cellSize);
   const marker = move % 2 === 0 ? 'X' : 'O';
   const x = boardLeft + c * cellSize + cellSize / 2;
   const y = boardTop + r * cellSize + cellSize / 2;
@@ -120,14 +122,11 @@ registerOnclick((x, y) => {
 
   let winner = checkWinner(lines, board)
 
-  let r = Math.floor((y - boardTop) / cellSize);
-  let c = Math.floor((x - boardLeft) / cellSize);
-
   // Only do anything if it's a legal move and the game isn't over.
   if (winner === null && underThree(r) && underThree(c) && board[r][c] === '') {
 
     // Draw the mark and record the move
-    mark(r,c)
+    mark()
 
     // Check if there's a winner now
     winner = checkWinner(lines, board);
--- a/assignments/refactor-tic-tac-toe/code.js
+++ b/assignments/refactor-tic-tac-toe/code.js
@@ -42,7 +42,7 @@ const wackyFunction = (lines, board, i) => {
   c = lines[i][1][1];
   const m1 = board[r][c];
   r = lines[i][2][0];
-  c = lines[i][2][2];
+  c = lines[i][2][1];
   const m2 = board[r][c];
   return [m0, m1, m2]
 }
--- a/assignments/refactor-tic-tac-toe/code.js
+++ b/assignments/refactor-tic-tac-toe/code.js
@@ -50,7 +50,7 @@ const board = [
   ['', '', ''],
 ];
 
-const wackyFunction = (lines,board) => {
+const wackyFunction = (lines,board,i) => {
   let r = lines[i][0][0];
   let c = lines[i][0][1];
   const m0 = board[r][c];
@@ -65,7 +65,7 @@ const wackyFunction = (lines,board) => {
 
 const checkWinner = (lines, board) => {
   for (let i = 0; i < lines.length; i++) {
-    m = wackyFunction(lines,board)
+    m = wackyFunction(lines,board,i)
     if (m[0] !== '' && m[0] === m[1] && m[0] === m[2]) {
       return lines[i]
     }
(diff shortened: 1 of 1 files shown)


assignments/refactor-tic-tac-toe/code.js
```diff
@@ -39,6 +50,31 @@ const board = [
   ['', '', ''],
 ];
 
+const checkWinner = (r, c, lines, board) => {
```
gigamonkey, line 53: Since you never actually use the values passed in for `r` and `c`, it would probably make more sense to make those local variables in the body of the function rather than arguments:

```javascript
const checkWinner = (lines, board) => {
  let r;
  let c;
```

assignments/refactor-tic-tac-toe/code.js
```diff
@@ -39,6 +50,31 @@ const board = [
...
+  for (let i = 0; i < lines.length; i++) {
+    r = lines[i][0][0];
+    c = lines[i][0][1];
+    const m0 = board[r][c];
```
gigamonkey, line 57: This chunk of three lines is basically duplicated twice below. That would be another thing to figure out how to factor out, maybe as a function or maybe rewriting these nine lines with loop. Or maybe both!

assignments/refactor-tic-tac-toe/code.js
```diff
@@ -72,26 +105,15 @@ registerOnclick((x, y) => {
...
-    c = lines[i][1][1];
-    const m1 = board[r][c];
-    r = lines[i][2][0];
-    c = lines[i][2][1];
-    const m2 = board[r][c];
-    if (m0 !== '' && m0 === m1 && m0 === m2) {
-      winner = lines[i];
-    }
+  if (checkWinner(r, c, lines, board)) {
+    winner = checkWinner(r, c, lines, board);
```
gigamonkey, line 109: Seems a bit gratuitous to call `checkWinner` twice here. You might need to tweak what `checkWinner` returns when it doesn't find a winner but you ought to be able to write just:

```javascript
winner = checkWinner(r, c, lines, board);
```

assignments/refactor-tic-tac-toe/code.js
```diff
@@ -104,47 +126,30 @@ registerOnclick((x, y) => {
...
-      c = lines[i][1][1];
-      const m1 = board[r][c];
-      r = lines[i][2][0];
-      c = lines[i][2][1];
-      const m2 = board[r][c];
-      if (m0 !== '' && m0 === m1 && m0 === m2) {
-        winner = lines[i];
-      }
+    if (checkWinner(r, c, lines, board)) {
+      winner = checkWinner(r, c, lines, board);
```
gigamonkey, line 130: Same comment as before.

assignments/refactor-tic-tac-toe/code.js
```diff
@@ -104,47 +126,30 @@ registerOnclick((x, y) => {
...
       const [r1, c1] = winner[0];
       const [r2, c2] = winner[winner.length - 1];
 
-      const x1 = boardLeft + c1 * cellSize + cellSize / 2;
-      const y1 = boardTop + r1 * cellSize + cellSize / 2;
-      const x2 = boardLeft + c2 * cellSize + cellSize / 2;
-      const y2 = boardTop + r2 * cellSize + cellSize / 2;
+      const location = (boardLocation, cr12) => {
+        let xOrY = boardLocation + cr12 * cellSize + cellSize / 2
+        return xOrY
```
gigamonkey, line 139: Whenever you have code in the pattern:

```javascript
let variable = something;
return variable;
```

you can just write:

```javascript
return something;
```

assignments/refactor-tic-tac-toe/code.js
```diff
@@ -39,6 +50,31 @@ const board = [
...
+      return lines[i]
+    }
+  }
+}
+
+const adjust = (firsts, seconds, adjustment, adjustees) => {
+  if (firsts[0] === firsts[1] || seconds[0] !== seconds[1]) {
+    const slope = y1 < y2 ? 1 : -1;
+    adjustees[0] -= (slope * adjustment);
+    adjustees[1] += (slope * adjustment);
```
gigamonkey, line 74: Modifying the value of `adjustees` can work though the way you actually call `adjust` it does not. But it's probably better to write this as a function that returns the adjusted values. 

```javascript
    return [
      adjustees[0] - (slope * adjustment),
      adjustees[1] + (slope * adjustment)
    ];
```

assignments/refactor-tic-tac-toe/code.js
```diff
@@ -104,47 +126,30 @@ registerOnclick((x, y) => {
...
-      }
+      adjust([y1, y2], [x1, x2], lineEndAdjustment, [adjX1, adjX2])
 
-      if (x1 === x2 || y1 !== y2) {
-        const slope = y1 < y2 ? 1 : -1;
-        adjY1 -= (slope * lineEndAdjustment);
-        adjY2 += (slope * lineEndAdjustment);
-      }
+
+      adjust([x1, x2], [y1, y2], lineEndAdjustment, [adjY1, adjY2])
```
gigamonkey, line 150: If you test your code, you may notice that it no longer actually adjusts the ends of the lines. That's because you're passing the array `[adjY1, adjY2]` as the fourth argument to this call and `adjust` is modifying that array but then you don't have a handle on the array so the adjustments are lost. (I.e. when you make that array you're just getting the *values* of `adjY1` and `adjY2`—changing the array has no effect on those variables.

gigamonkey (changes requested):
This is a very good start. I left some comments for more things for you to try.

assignments/refactor-tic-tac-toe/code.js
```diff
@@ -39,6 +24,49 @@ const board = [
...
+  const marker = move % 2 === 0 ? 'X' : 'O';
+  const x = boardLeft + c * cellSize + cellSize / 2;
+  const y = boardTop + r * cellSize + cellSize / 2;
+  const nudge = marker === 'O' ? cellSize / 9 : cellSize / 19;
+  drawText(marker, x - (fontSize * 0.3 + nudge), y + fontSize * 0.3, 'black', fontSize);
+  board[r][c] = marker;
+  move++;
+}
+
+const wackyFunction = (lines, board, i) => {
```
gigamonkey, line 37: Nice. This is a good function to break out though obviously at some point it's worth coming up with a more descriptive name. I think I had a similar function in one of my refactors and called it `exractLine`.

assignments/refactor-tic-tac-toe/code.js
```diff
@@ -60,91 +88,43 @@ const x1 = boardLeft + cellSize;
...
       const [r1, c1] = winner[0];
       const [r2, c2] = winner[winner.length - 1];
 
-      const x1 = boardLeft + c1 * cellSize + cellSize / 2;
-      const y1 = boardTop + r1 * cellSize + cellSize / 2;
-      const x2 = boardLeft + c2 * cellSize + cellSize / 2;
-      const y2 = boardTop + r2 * cellSize + cellSize / 2;
+      const location = (boardLocation, cr12) => {
+        return boardLocation + cr12 * cellSize + cellSize / 2
+      }
```
gigamonkey, line 115: This is another good function. Because it's so short it doesn't matter that you defined it as a nested function but I'd probably make it a top-level function just to make it immediately obvious that it doesn't depend on anything else in the enclosing function.

gigamonkey (approved):
Great job! I think your're ready for Friday. If you want, tomorrow you can start thinking about what you might want to do for your next project. (We'll be mixing some new material with work on projects for the rest of they year, pretty much.)

PR #51, opened 2023-02-23
decided to fix code, works (used example from you)

elijahsur:


gigamonkey (approved):
Nice. Glad to see you came back to it.

PR #53, opened 2023-02-27
Assignments/hof implementation

elijahsur:
not done!

### DIFF
--- a/assignments/hof-implementation/code.js
+++ b/assignments/hof-implementation/code.js
@@ -3,10 +3,11 @@
 const filter = (predicate, array) => {
   let nArray = []
   for (let i = 0; i < array.length; i++) {
-    if (predicate(n)) {
+    if (predicate(i)) {
       nArray.push[array[i]]
       }
   }
+  return nArray
 };
 
 const map = (fn, array) => {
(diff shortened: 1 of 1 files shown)


gigamonkey (changes requested):
Okay. A good start. Keep on going! (I assume `flatMap` is not passing the tests; I'll let you debug that yourself.)

assignments/hof-implementation/code.js
```diff
@@ -1,19 +1,51 @@
...
+  }
+  return nArray
 };
 
 const flatMap = (fn, array) => {
+  let nArray = []
+  for (let i = 1; i < array.length + 1; i++) {
+    for (let j = 0; j < fn(i).length; j++) {
+      nArray.push(fn(i)[j])
+    }
```
gigamonkey, line 24: This works but for a couple reasons (efficiency being one) it's probably better to only call `fn` once per element of the array. This is going to call it like 2 times the number of items in the array`fn(i)` returns for every `i`. I.e. something like this:

```javascript
    const subArray = fn(i);
    for (let j = 0; j < subArray.length; j++) {
      nArray.push(subArray[j])
    }
```

gigamonkey (approved):
Looks good. A couple style points (though the one about `flatMap` is a bit more important than totally cosmetic.) Anyway, you can merge this and move on to the other assignment if you haven't yet.

PR #54, opened 2023-02-28
Assignments/array hofs

elijahsur:
not sure whats going on with the last function


assignments/array-hofs/code.js
```diff
@@ -0,0 +1,16 @@
+const evens = (a) => a.filter((a) => a % 2 === 0)
+const odds = (a) => a.filter((a) => a % -2||2 === 1)
```
gigamonkey, line 2: Wow. I'm not quite sure how you arrived at this test for something being odd. It does work but only via some very convoluted paths. I assume you went down this path after maybe trying `a % 2 === 1` or something that didn't work quite right for negative values of `a`. The reason for that is kind of annoying but a simpler way to get an odd test that works for both positive and negative is just to reverse the evens test: `a % 2 !== 0`.

assignments/array-hofs/code.js
```diff
@@ -0,0 +1,16 @@
...
+const big = (a) => a.filter((a) => a > 100)
+const names = (a) => a.flatMap((a) => a["name"])
```
gigamonkey, line 4: Note you can use the `a.name` syntax here and probably should when you can.

assignments/array-hofs/code.js
```diff
@@ -0,0 +1,16 @@
...
+const grades = (a) => a.flatMap((a) => a["grade"])
```
gigamonkey, line 5: Likewise, `a.grade` is preferable though this also works.

assignments/array-hofs/code.js
```diff
@@ -0,0 +1,16 @@
...
+const pairs = (a) => a.map((a) => [a, a])
+const averageGrade = (a) => a.flatMap((a) => a["grade"]).reduce((tot, a) => tot + a) / a.length
```
gigamonkey, line 7: No need to use `flatMap` here though I guess it works because Javascript is a bit sloppy and can deal with a single item being returned by the function. But since you know you're always returning a single grade, `map` would suffice and would be preferable so we don't wonder what's up with the `flatMap`.

assignments/array-hofs/code.js
```diff
@@ -0,0 +1,16 @@
...
+const flatpairs = (a) => a.flatMap((a) => [a, a])
+const allEven = (a) => a.every((a) => a % 2 === 0)
+const someEven = (a) => a.some((a) => a % 2 === 0)
+const lengthOfNames = (a) => {
+  let uppercase = ["A","B","C","D","E","F","G","H","I","J","K","L","M","N","O","P","Q","R","S","T","U","V","W","X","Y","Z"]
+  a.map((a) => a[0])
+  return a
+}
+lengthOfNames(['Fred','Barney','car','Wilma','rock','Betty','dinosaur'])
```
gigamonkey, line 16: If you're still stuck on this,ask me tomorrow and I can help.

gigamonkey (changes requested):
Just the last one, as you say. I can help you with it tomorrow.

assignments/array-hofs/code.js
```diff
@@ -0,0 +1,14 @@
...
+const big = (a) => a.filter((a) => a > 100)
+const names = (a) => a.map((a) => a.name)
+const grades = (a) => a.map((a) => a.grade)
+const pairs = (a) => a.map((a) => [a, a])
+const averageGrade = (a) => a.flatMap((a) => a.grade).reduce((tot, a) => tot + a) / a.length
+const flatpairs = (a) => a.flatMap((a) => [a, a])
+const allEven = (a) => a.every((a) => a % 2 === 0)
+const someEven = (a) => a.some((a) => a % 2 === 0)
+const lengthOfNames = (a) => {
+  let length = a.filter((n) => n[0].toUpperCase() === n[0])
```
gigamonkey, line 12: This works though I'd probably pick a variable name that names what the value actually is. Since after the `filter` you have an array of just the names I'd probably call it `names` rather than `length`. Or I might just chain everything together and skip the variable altogether:

```javascript
  return a.filter((n) => n[0].toUpperCase() === n[0]).map((a) => a.length);
```

gigamonkey (approved):
Nice work. You should be thinking about/getting started on your next project. Happy to discuss what you want to do.

PR #55, opened 2023-03-03
Projects/vender algorithm

elijahsur:


projects/vender-algorithm/public/scripts/algorithm.js
```diff
@@ -0,0 +1,43 @@
...
+        cityArray.push([Math.floor(Math.random() * boardSize), Math.floor(Math.random() * boardSize)])
+    }
+    return cityArray
+}
+// judges based on if it goes to any city multiple times
+const fitness = (path) => {
+    let fitness = 0
+    for (let i = 0; i < path.length; i++) {
+        if (!path.every((n) => path[i] === n)) {
+            fitness + 1 / path.length
```
gigamonkey, line 16: Note that this is just an expression that computes a value but you're not doing anything with the value, i.e. you're not assigning it back to `fitness` which is kind of what I'd expect.

projects/vender-algorithm/public/scripts/algorithm.js
```diff
@@ -0,0 +1,43 @@
...
+
+// makes random cities on a grid, represented with an x and y axis
+const makeCities = (nCities, boardSize) => {
+    let cityArray = []
+    for (let i = 0; i < nCities; i++) {
+        cityArray.push([Math.floor(Math.random() * boardSize), Math.floor(Math.random() * boardSize)])
+    }
+    return cityArray
+}
+// judges based on if it goes to any city multiple times
```
gigamonkey, line 11: The real measure of fitness is the length of the path (in terms of distance traveled in the x/y plane). You'll probably want to arrange things so that the "genome" of possible solutions can only encode paths that satisfy the only-once criteria. E.g. if your organisms are just lists of cites with no duplicates then they will all trivially satisfy the only-once criteria. (On the other hand you'll then need to exercise some care in your crossing and mutating to ensure that you don't break that structure. E.g. you can't just blindly change one city to another in a mutation because you'd likely remove one city and duplicate another. We can talk about different ways you might handle this.)

projects/vender-algorithm/public/scripts/algorithm.js
```diff
@@ -0,0 +1,43 @@
+// travelling salesman problem code
+
+// makes random cities on a grid, represented with an x and y axis
+const makeCities = (nCities, boardSize) => {
+    let cityArray = []
+    for (let i = 0; i < nCities; i++) {
+        cityArray.push([Math.floor(Math.random() * boardSize), Math.floor(Math.random() * boardSize)])
```
gigamonkey, line 7: Might be worth breaking out a `randomNumber(max)` function that captures the `Math.floor(Math.random() * value)` expression, just because that's kinda a mouthful.

gigamonkey (approved):
Good start.

projects/vender-algorithm/public/scripts/algorithm.js
```diff
@@ -0,0 +1,54 @@
...
+// creates the first batch of paths
+const firstBatch = (cityList, batchAmount) => {
+    let batch = []
+    for (let i = 0; i < batchAmount; i++) {
+        let indBatch = [...cityList]
+        for (let j = 0; j < cityList.length; j++) {
+            let random = cityList[maxRandom(cityList.length)]
+            indBatch.splice(indBatch.indexOf(random), 1)
+            indBatch.push(random)
+        }
```
gigamonkey, line 37: Might be worth breaking these lines out to a function that takes an array and returns a shuffled copy of the array. (Then later you may want to look up other ways to shuffle an array--what you have here is quite close to the canonical shuffling algorithm, the Fisher-Yates shuffle but I think not quite right in a way that might mean it doesn't necessarily produce all permutations (orderings) with equal probability. But it's pretty darn close!)

PR #57, opened 2023-03-14
evolve (maybe) not working, else good

elijahsur:
im not sure if it is working as intended, or if its just a bug, but it gets stuck at a certain path after a while of running

### DIFF
--- a/projects/vender-algorithm/public/scripts/algorithm.js
+++ b/projects/vender-algorithm/public/scripts/algorithm.js
@@ -2,7 +2,7 @@
 
 // shuffles an array, switching positions of parts of the array
 const arrayShuffler = (array) => {
-    let indBatch = [...array];
+    let indBatch = [...array]
     for (let j = 0; j < array.length; j++) {
         let random = array[maxRandom(array.length)]
         indBatch.splice(indBatch.indexOf(random), 1)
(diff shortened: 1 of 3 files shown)


projects/vender-algorithm/public/scripts/algorithm.js
```diff
@@ -64,37 +64,45 @@ const updateFitness = (paths) => {
...
+    paths.sort((b, a) => a.fitness - b.fitness);
     return paths.slice(Math.floor(paths.length / 2), paths.length)
 };
 
 const evolve = (paths, batchN) => {
-        for (let i = 1; i < paths.length; i++) {
-            let nPath = {'path': paths[i].path, 'fitness': paths[i].fitness}
+    let limit = paths.length
+    for (let i = 1; i < batchN - limit; i++) {
+        let newPath = { 'path': paths[i - 1].path, 'fitness': paths[i - 1].fitness }
```
gigamonkey, line 74: You may be having a similar problem as before when you needed to make a copy of the array. `newPath.path` is going to be the same path array as `paths[i - 1].path`. So later on when you assign to `newPath.path[switchedPart.index]` you're changing the original path too. You can fix this here by the same technique you used before:

```javascript
        let newPath = { 'path': [...paths[i - 1].path], 'fitness': paths[i - 1].fitness }
```

Though I might instead write this whole thing slightly differently where I start with something like:

```javascript
        let newPath = { 'path': [], 'fitness': 0 }
```

And then fill in `newPath.path` by taking the appropriate bits from the appropriate places and then scoring the fitness at the end. (Actually, come to think of it, I'd probably write a function that just takes two paths and returns new path via your crossing mechanism (and without modifying the arguments in any way) and then make a new critter out of that path and it's computed fitness.

gigamonkey (commented):
You didn't ask for review on this yet but I had a thought about the question asked the other day in class.

PR #58, opened 2023-03-21
read comment on drawPath script.js

elijahsur:


projects/vender-algorithm/public/scripts/script.js
```diff
@@ -8,16 +8,64 @@ let boardSize = 500
...
+  // }
+  can.beginPath()
+  can.strokeStyle = "black"
+  can.moveTo(pandf.path[0].x, pandf.path[0].y)
+  for (let i = 1; i < pandf.path.length; i++) {
+    can.lineTo(pandf.path[i].x, pandf.path[i].y)
+    can.beginPath()
+    can.ellipse(pandf.path[i].x, pandf.path[i].y, 5, 5, 0, 0, 360)
+    can.fill()
+    can.stroke
```
gigamonkey, line 59: You're missing the `()`s on this call to `stroke`.

projects/vender-algorithm/public/scripts/script.js
```diff
@@ -8,16 +8,64 @@ let boardSize = 500
...
+  }
```
gigamonkey, line 60: I think you may need to break drawing the lines and the ellipses into separate paths because both `stroke` and `fill` both close the current path. I also don't think you can meaningfully nest paths so after you call `beginPath()` you need to call `stroke()` or `fill()` before the next `beginPath()`. Here's a sample of how I did it as a test:

```javascript
const canvas = document.getElementById('screen');
const ctx = canvas.getContext('2d');
const width = canvas.width;
const height = canvas.height;

const r = 10;

const cities = Array(50).fill().map(() => ({
  x: r + Math.random() * (width - 2 * r) ,
  y: r + Math.random() * (height - 2 * r)
}));

const drawLines = (cities) => {
  ctx.strokeStyle = 'blue';
  ctx.beginPath();
  ctx.moveTo(cities[0].x, cities[0].y);
  for (let i = 1; i < cities.length; i++) {
    ctx.lineTo(cities[i].x, cities[i].y);
  }
  ctx.lineTo(cities[0].x, cities[0].y);
  ctx.stroke()
};

const drawCircles = (cities) => {
  for (let i = 0; i < cities.length; i++) {
    ctx.fillStyle = 'red';
    ctx.beginPath();
    ctx.ellipse(cities[i].x, cities[i].y, r, r, 0, 0, 2 * Math.PI);
    ctx.fill();
  }
};

const draw = (cities) => {
  drawLines(cities);
  drawCircles(cities);
};

draw(cities);
```

gigamonkey (approved):
Left some sample code for drawing lines and circles.

PR #59, opened 2023-03-28
not sure how to put text in new button

elijahsur:
changed the button to be a different shape, and at the top of the screen.
how would i put text in the middle of the button?
another option i was thinking of was to make the entire canvas a button, and it generates when you click it.
how could i do that?

### DIFF
--- a/projects/vender-algorithm/public/index.html
+++ b/projects/vender-algorithm/public/index.html
@@ -9,11 +9,11 @@
 </head>
 <body>
   <h1>vender algorithm</h1>
+  <input id="generate" type="button"</input>
   <canvas></canvas>
   <p> </p>
   <div id="path"><fieldset>
     <legend><p>path</p></legend>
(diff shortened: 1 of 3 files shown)


projects/vender-algorithm/public/index.html
```diff
@@ -9,11 +9,11 @@
 </head>
 <body>
   <h1>vender algorithm</h1>
+  <input id="generate" type="button"</input>
```
gigamonkey, line 12: You broke the syntax a bit here, you're missing the `>` on the opening tag.

gigamonkey (commented):
I'm not sure I understood your question but if you want to make a button with specific text, the modern way is with the `<button>` element rather than `<input type="button">`. The `<button>` element can have whatever you want as it's content, e.g.

```html
<button>This is some text with <b>other</b> markup.</button>
```

PR #60, opened 2023-03-29
pretty much done, suggest anything to change

elijahsur:


gigamonkey (commented):
Sorry, I got a bit behind on reviewing. This is great. I haven't actually had a chance to go over the code with a fine tooth comb but one thing I notice is that I maybe never quite explained what I meant by a "brute force" solver. It seems that what you implemented for that was something that repeatedly generates random solutions and keeps the best one. 

What I actually meant by "brute force" is something that systematically tries all the possibilities and keeps the best one. I.e. if you had a problem with four cities, A, B, C, and D you would try the paths:

- ABCD
- ABDC
- ACBD
- ACDB
- ADBC
- ADCB

And report which one was shortest. I.e. a brute force solver will definitely find the absolute best solution, assuming it finishes. The problem with the traveling salesman problem (and many others) is that brute force solutions are useful because the number of possible solutions grows so quickly as the size of the problem grows that very soon it will take more than the lifetime of the universe to check them all.

But the idea here was that if you had a brute force solver you could test your genetic algorithm on small problems and see if it actually finds the best possible solution. If it can do that for small problems that you can have some confidence that it basically works and will likely find at least good solutions to bigger problems.

PR #61, opened 2023-03-30
Projects/image algorithm

elijahsur:


projects/image-algorithm/public/script.js
```diff
@@ -0,0 +1,31 @@
...
+
+let ctx = canvas.getContext("2d");
+
+//specific max amount, then generates random numbers
+const maxRandom = (max) => {
+    return Math.floor(Math.random() * max)
+};
+
+//draws the first batch of random circles around the canvas
+const randomCircles = () => {
```
gigamonkey, line 19: Note that the thing I was talking about was random triangles. Triangles have the advantage of coming in more than one shape and overlapping in lots of interesting ways.

projects/image-algorithm/public/index.html
```diff
@@ -0,0 +1,24 @@
...
+  <p> </p>
+  <fieldset>
+    <legend><p>generate</p></legend>
+    <input type="button" id="generate"></input>
+  </fieldset>
+  <fieldset>
+    <legend><p>circle count</p></legend>
+    <input type="text" id="circles"></input>
+  </fieldset>
+  <img src="/Mona_Lisa.jpeg" id="mona-lisa">
```
gigamonkey, line 22: Ah, so one of the things you'll need to figure out (or I can show you) is how to load an image into a Canvas using the drawing APIs because that is the best way (that I know of) to get access to the image data we were talking about. So what you should probably have in your HTML are two `<canvas>` elements, one into which you'll load the reference image and another where you can draw your critters.

gigamonkey (approved):
Good start anyway.

PR #62, opened 2023-04-10
Assignments/recursion

elijahsur:


assignments/recursion/code.js
```diff
@@ -0,0 +1,31 @@
...
+  }
+}
+
+const fibonacci = (n) => {
+  if (n === 0) {
+    return 0
+  } else {
+    return n + fibonacci(n - 1)
+  }
+}
```
gigamonkey, line 23: This doesn't pass the test cases, right? I updated the description of the problem in the assignment in a way that may give you a useful clue about how to fix it.

assignments/recursion/code.js
```diff
@@ -0,0 +1,31 @@
...
+
+const gcd = (a, b) => {
+  if (b === 0) {
+    return a
+  } else {
+    return b - a % b
```
gigamonkey, line 29: Remember that the recursive case always has a recursive call to the function.

gigamonkey (changes requested):
Good start! Keep going.

assignments/recursion/code.js
```diff
@@ -0,0 +1,71 @@
...
+  if (a.length !== 0) {
+  if (a[0] === v) {
+    return true
+  } else {
+    return search(a.slice(1), v)
+  }
+  } else {
+    return false
+  }
+}
```
gigamonkey, line 55: This is right, though you should run "Format document" to fix up the indentation. My preference is to write the base case first so I might write this as:

```javascript
const search = (a, v) => {
  if (a.length === 0) {
    return false;
  } else {
    if (a[0] === v) {
      return true
    } else {
      return search(a.slice(1), v)
    }
  }
}
```

In this case I'm violating the advice I gave about about `else if`s because I want to emphasize first the main branch between the base case and the recursive case and then the second `if`/`else` is just the implementation of the recursive case.

I might then simplify the recursive case to not require another level of `if`/`else` by writing a boolean expression:

```javascript
const search = (a, v) => {
  if (a.length === 0) {
    return false;
  } else {
    return a[0] === v || search(a.slice(1), v);
  }
}
```

This takes advantage of the fact that `||` is "short-circuiting"—if the first operand (`a[0] === v` in this case) evaluates to `true` then the whole `||` expression will be true so the second operand, the recursive call to `search`, doesn't need to be evaluated.

gigamonkey (changes requested):
Nice work. One more to go!

assignments/recursion/code.js
```diff
@@ -0,0 +1,82 @@
...
+
+const treeMap = (t, f) => {
+  if (isLeaf(t)) {
+    return f(t)
+  } else {
+    return { 'left': treeMap(t.left, f), 'right': treeMap(t.right, f) }
+  }
+}
+
+const change = (amt, coins) => {
```
gigamonkey, line 72: Nice. Now to just wrap your head around why it works! ;-)

gigamonkey (approved):
Very nice work!

PR #63, opened 2023-04-13
got image to load and rgba information

elijahsur:


projects/image-algorithm/public/script.js
```diff
@@ -25,6 +25,10 @@ const maxRandom = (max) => {
 
 //draws the first batch of random triangles around the canvas
 const randomTriangles = () => {
+    let testx = maxRandom(picture.width)
+    let testy = maxRandom(picture.height)
+    const data = cdy.getImageData(testx, testy, 1, 1).data
```
gigamonkey, line 30: This works though you might actually want, later on, I think, to do something like `cdy.getImageData(0, 0, picture.width, picture.height).data` to get all the image data at once and then access the array to get the bits you want for any given pixel. (I haven't actually checked but my intuition is that getting the image data all at once is going to be cheaper than getting it via `getImageData` a pixel at a time.) But this is a great step forward, making sure the API is working the way you expect.

gigamonkey (approved):
A good step forward!

gigamonkey (changes requested):
So the main thing, as we talked about, is to make some kind of data structure you can use to determine what triangles to draw where on a canvas but also can serve as the "dna" of your organisms that you can then score the fitness of, cross, and mutate to make new dna.

I'd probably go with something like an array of objects, maybe something like this:

```javasrcipt
[
  { 
    points: [
      { x: 0, y: 0}, 
      { x: 13, y: 42 }, 
      { x: 49, y: 200 }, 
    ],
    color: { r: 127, g: 35, b: 255, a: 0.73 },
  },
  {
    points: [
      { x: 10, y: 0},
      { x: 45, y: 21 },
      { x: 59, y: 120 },
    ]
    color: { r: 227, g: 69, b: 25, a: 0.21 },
  },
   // and so on ...
]
```

projects/image-algorithm/public/script.js
```diff
@@ -81,9 +80,40 @@ const fitness = () => {
...
+        let newTri = batches[i][random]
+        let oldTri = batches[i - 1][random]
+        batches[i - 1][random] = newTri
+        batches[i][random] = oldTri
+    }
+    return batches
 }
 
-document.querySelector('#generate').onclick = () => batchFitness();
+const drawBatch = (batch) => {
```
gigamonkey, line 98: I'm a bit confused about the naming here. "batch" to me implies a set of similar things, like a batch of cookies or a batch of recruits. Yet this function seems to be dealing with just one item, namely one triangle. Why isn't this called 'drawTriangle' and the argument `triangle`?

gigamonkey (approved):
This looks sensible to me though I do have some questions about your naming, noted below.

gigamonkey:
Oh hey, I started working on my own version of this and realized that you'll probably need a way to convert RGBA values like you get from `getImageData` into plain RGB values since each pixel is only one color and it doesn't really matter if two pixels have different RGBA values if the effective RGB values are the same. I had previously written these functions which you might find useful. I'm pretty sure I tested it thoroughly when I wrote it so it should be right.

```javascript
/*
 * Translate value at a given alpha to the corresponding value at full opacity.
 * Loosely based on code from https://stackoverflow.com/a/11615135.
 */
const opaquify = (v, a) => Math.round((v * a) / 255 + (255 - a));

/*
 * Translate rgba values we get from ImageData into RGB triples.
 */
const toRGB = (r, g, b, a) => [r, g, b].map((v) => opaquify(v, a));
 ```

PR #64, opened 2023-04-17
Assignments/recursion 2

elijahsur:


assignments/recursion-2/code.js
```diff
@@ -0,0 +1,75 @@
...
+}
+
+const maximum = (ary) => {
+  if (ary.length === 0) {
+    return -Infinity
+  } else {
+    if (ary[0] < ary[1]) {
+      return maximum(ary.slice(1))
+    } else {
+      return maximum(ary[0] + ary.slice(1))
```
gigamonkey, line 72: Don't think that works, does it? You could make the array you want either by:

```javascript
[ary[0]].concat(ary.slice(1))
```

But maybe better is to remember `Math.max()` which can take two numbers and return the larger of the two. E.g. `Math.max(10, 20)` return `20`. (Technically it can be used with more than two arguments but for now assume it has to be two.) Try finding a way using that method.

gigamonkey (changes requested):
Very nice work. One little glitch with `maximum` to fix and then `every` and `some` but you clearly get how recursion works.

assignments/recursion-2/code.js
```diff
@@ -0,0 +1,99 @@
...
+
+const maximum = (ary) => {
+  if (ary.length === 0) {
+    return -Infinity
+  } else {
+    if (ary[0] < ary[1]) {
+      return maximum(ary.slice(1))
+    } else {
+      return Math.max(ary[0], maximum(ary.slice(1)))
+    }
```
gigamonkey, line 73: You can further simplify this to just:

```javascript
return Math.max(ary[0], maximum(ary.slice(1)));
```

And you actually kind of need to because `ary[1]` isn't necessarily defined; if `ary` is only one element long it won't be. But `Math.max` handles comparing to `-Infinity` with no problem.

gigamonkey (approved):
Looks good. One nit about `maximum` that you should make sure you understand.

PR #65, opened 2023-04-19
Assignments/recursion 3

elijahsur:


assignments/recursion-3/code.js
```diff
@@ -0,0 +1,77 @@
...
+    return 2
+  } else if (n === 1) {
+    return 1
+  } else {
+    return lucas(n - 2) + lucas(n - 1)
+  }
+}
+
+const isAscending = (ary) => {
+  if (ary.length === 1) {
```
gigamonkey, line 28: This would be slightly better with `ary.length < 2` as the condition because it handles the empty array and also ensures that we can safely access both `ary[0]` and `ary[1]` in the recursive case.

assignments/recursion-3/code.js
```diff
@@ -0,0 +1,77 @@
...
+    return true
+  } else if (ary[0] <= ary[1]) {
+    return isAscending(ary.slice(1))
+  } else {
+    return false
+  }
+}
+
+const isDescending = (ary) => {
+  if (ary.length === 1) {
```
gigamonkey, line 38: Same comment, `ary.length < 2` would be better.

assignments/recursion-3/code.js
```diff
@@ -0,0 +1,77 @@
...
+    return 0
+  } else {
+    return sumNested(aryN[0]) + sumNested(aryN.slice(1))
+  }
+}
+
+const searchNested = (ary, n) => {
+  if (ary === n) {
+    return true
+  } else if (ary.length === 0) {
```
gigamonkey, line 60: This isn't quite right (though it's possible it passes the test cases because Javascript is sloppy) because you haven't established that `ary` is an array at this point. If it's a number then `ary.length is undefined` which is, as it happens, not `=== 0` so this works but not really for the right reason. And in a stricter language this would be an error because numbers don't have a length property. Better to put the check for `isNumber` early and return for those cases and then the remaining cases can safely assume `ary` is in fact an array.

assignments/recursion-3/code.js
```diff
@@ -0,0 +1,77 @@
...
+    return true
+  } else if (isNumber(ary)) {
+    return false
+  } else {return searchNested(ary.slice(1), n)}
+}
+
+const searchNested2 = (ary, n) => {
+  if (isNumber(ary)) {
+    if (ary === n) { return true } else {
+      return searchNested(ary.slice(1), n)
```
gigamonkey, line 72: This can''t be quite right because if we're in the `if (isNumber(ary))` branch, then we know `ary` is a number and therefore it doesn't make sense to call `slice` on it.

Check out the posted answer if you want to see how I'd structure it: https://github.com/bhs-intro-to-programming/answers/blob/main/assignments/recursion-3/code.js#L53

gigamonkey (changes requested):
Looks good. You should take a crack at evaluate—it's not nearly as hard a `change`; I think you can do it.

PR #68, opened 2023-04-27
Projects/image algorithm

elijahsur:


projects/image-algorithm/public/script.js
```diff
@@ -116,22 +107,35 @@ const batchFitness = () => {
...
 }
 
 const drawTriangle = (triangle) => {
-    ctx.globalAlpha = 0.5
+    ctx.globalAlpha = 1
+    ctx.beginPath()
+    ctx.fillStyle = 'red'//`rgba(${triangle.color.r}, ${triangle.color.g}, ${triangle.color.b}, ${triangle.color.a})`
+    console.log(ctx.fillStyle)
+    ctx.moveTo(triangle.points[0].x, triangle.points[0].y)
+    for (let i = 1; i < 2; i++) {
```
gigamonkey, line 125: Oh no! I see the problem. How many times does this loop run? (FWIW, I saw this by running your code adding some console logging of what coordinates it was `moveTo`ing to and then `lineTo`'ing to and noticed that it wasn't calling `lineTo` enough times.

I'll note that this would have been avoided by using `triangle.points.length` instead of `2` in the loop condition.

gigamonkey (changes requested):
Found the bug. Gave you a pretty strong hint of where it is.

projects/image-algorithm/public/script.js
```diff
@@ -81,13 +92,15 @@ const fitness = () => {
             // array to get the specific pixels.
             let genPixel = ctx.getImageData(j, i, 1, 1).data
             let picPixel = cdy.getImageData(j, i, 1, 1).data
-            totFitness = totFitness + (picPixel[0] - genPixel[0] + picPixel[1] - genPixel[1] + picPixel[2] - genPixel[2] + picPixel[3] - genPixel[3])
+            let rgbGenPixel = toRGB(genPixel)
+            let rgbPicPixel = toRGB(picPixel)
```
gigamonkey, line 96: `toRGB` expect four arguments and you're only passing it one. Since `genPixel` and `picPixel` are four-element arrays, you can pass them with the "spread" syntax: `toRGB(...genPixel)` which is a shorthand for `toRGB(genPixel[0], genPixel[1], genPixel[2], genPixel[3])`. Alternatively, you're  free to change `toRGB` to expect an array rather than four separate arguments.

gigamonkey (changes requested):
I mostly looked at the place you were using `toRGB` and left a comment about what I suspect the problem is.

projects/image-algorithm/public/script.js
```diff
@@ -78,24 +77,22 @@ const opaquify = (v, a) => Math.round((v * a) / 255 + (255 - a));
...
+    let imgData = cdy.getImageData(0, 0, picture.width, picture.height).data
+    console.log({'drawData length' : drawData.length, 'img size' : picture.width * picture.height, 'size difference' : drawData.length - picture.width * picture.height * 4})
+    for (let i = 0; i < (drawData.length); i++) {
+            let drawPixel = drawData.slice(0, 4)
+            drawData = drawData.slice(4)
+            let imgPixel = imgData.slice(0,4)
+            imgData = imgData.slice(4)
+            let rgbDraw = toRGB(...drawPixel)
+            let rgbIMG = toRGB(...imgPixel)
+            totFitness = totFitness + (rgbIMG[0] - rgbDraw[0] + rgbIMG[1] - rgbDraw[1] + rgbIMG[2] - rgbDraw[2])
```
gigamonkey, line 92: You might want to think about this fitness function. They way this is written, having a pixel with, for instance, a red value that it 10 less than the reference image will add to the fitness score while having a red value that is 10 greater will decrease the fitness score. What you really want to measure is the distance from the desired value. So `-Math.abs(rgbIMG[0] - rgbDraw[0])` or something, i.e. the greater the distance between the two values the more fitness is decreased. Or if you add the squares of the differences, at the end you can return `Math.sqrt(totalFitness)` and get the Euclidean distance between the two images in a many-dimensional space. (That's how I did it.)

gigamonkey (approved):
Some hints about how to possibly improve the performance. Also a note about your fitness function.

projects/image-algorithm/public/script.js
```diff
@@ -82,17 +82,12 @@ const fitness = () => {
...
+    for (let i = 0; i < (drawData.length); i += 4) {
+            let drawPixel = drawData.slice(i, i + 4)
+            let imgPixel = imgData.slice(i, i + 4)
             let rgbDraw = toRGB(...drawPixel)
             let rgbIMG = toRGB(...imgPixel)
-            totFitness = totFitness + (rgbIMG[0] - rgbDraw[0] + rgbIMG[1] - rgbDraw[1] + rgbIMG[2] - rgbDraw[2])
-            if (i % 100 === 0) {
-                console.log([i, rgbDraw])
-            }
+            totFitness = totFitness + Math.sqrt((rgbIMG[0] - rgbDraw[0])^2 + (rgbIMG[1] - rgbDraw[1])^2 + (rgbIMG[2] - rgbDraw[2])^2)
```
gigamonkey: Unfortunately `^` *is* an operator in Javascript but it is *not* the exponentiation operator which is spelled `**`. (`^` is he bitwise XOR operator so it treats its two operands as numbers and exclusive-or's them bit by bit. So anything^2 ether toggles the bit in the twos place of the number in binary on or off, either adding two or subtracting two.

gigamonkey (changes requested):
Problem with exponentiation. See below.

PR #69, opened 2023-05-09
doesnt generate to match

elijahsur:


projects/image-algorithm/public/script.js
```diff
@@ -47,7 +47,7 @@ const randomTriangles = () => {
         let g = randomRGBA()
         let b = randomRGBA()
         let a = randomRGBA()
-        ctx.fillStyle = `rgba(${r}, ${g}, ${b}, ${a})`
+        ctx.fillStyle = `rgba(${r}, ${g}, ${b}, ${a*255})`
```
gigamonkey, line 50: Did I type that wrong? I think it should be `a/255` so we get a number from 0.0 to 1.0 if `a` is in the range 0-255.

projects/image-algorithm/public/script.js
```diff
@@ -111,18 +111,64 @@ const batchFitness = (batch) => {
...
+const allChange = (triangle) => {
+    let nPoints = []
+    for (let i = 0; i < triangle.points.length; i++) {
+        nPoints.push({'x' : triangle.points[i].x + maxRandom(100) - 50, 'y' :  triangle.points[i].y + maxRandom(100) - 50})
+    }
+    return {'points' : nPoints, 'color' : colorChange(triangle.color)}
+}
+
+const colorChange = (color) => {
+    let changeFactor = maxRandom(25)
```
gigamonkey, line 154: Probably want to make this a number in a range from some negative to some positive value.

projects/image-algorithm/public/script.js
```diff
@@ -111,18 +111,64 @@ const batchFitness = (batch) => {
...
+    if (!color.r + changeFactor > 255) {
```
gigamonkey, line 155: I'm pretty sure the precedence of this is not what you want. `!` has quite high precedence so this is parsed like:

```javascript
((!color.r) + changeFactor) > 255
```

where `!color.r` first converts `color.r` to a boolean (`false` if the value is `0` and `true` otherwise). Then boolean is converted back to a number (`0` or `1`) and added to `changeFactor` yielding a number that is at most 26 and therefore not `> 255` so this expression is always `false`.

You probably want:

```javascript
if (!(color.r + changeFactor > 255)) {
```

Or:

```javascript
if (color.r + changeFactor <= 255) {
```

But if you make `changeFactor` possibly negative then you want to clamp the value in the range 0-255. I usually write a function like:

```javascript
const clamp = (n, min, max) => Math.min(max, Math.max(min, n));
```

Then you can do things like `color.r = clamp(color.r + changeFactor, 0, 255)`.

projects/image-algorithm/public/script.js
```diff
@@ -146,7 +192,7 @@ const run = (batchcount, limit) => {
         console.log(batches.length)
         batches = rank(mBatches)
         console.log('ranked')
-        console.log([batches.length, batches[0].fitness])
+        console.log([batches.length, Math.trunc(batches[0].fitness * 1000000) + '% matching'])
```
gigamonkey, line 195: If you're trying to limit the number of digits printed, you can use `toFixed` a method on numbers. E.g. `batches[0].fitness.toFixed(3)` will turn the number into a string with three places after the decimal point.

gigamonkey (changes requested):
This is some tricky stuff to debug! The main thing I think you need to do is figure out how to make sure things are working the way you expect. For instance, I'm pretty sure due to the precedence error with `!` your mutation was never changing the color of the triangles. Which might have been hard to see if you had too many triangles. So you can strip things down so you're only generating one triangle and drawing it and then you can see whether it is changing color. When you notice it is not then obviously you'd have to figure out why which might be tricky if you didn't know about the precedence of `!` but if you put some console logging in where the color was changed and saw that it never happened you'd probably eventually puzzle out what was going on. This is in fact probably the most important skill in programming because debugging is a constant task, no matter how experienced you are. (The main difference is with more experience you get better at debugging quickly. And you do have more experience to draw on of the kind of things, e.g. precedence errors, that can go wrong.)

projects/image-algorithm/public/script.js
```diff
@@ -157,5 +195,26 @@ const rank = (batches) => {
...
+        let child = { 'fitness': 0, 'triangles': mutate(mix(triAndFit.triangles), chance) }
+        offspring.push(batchFitness(child))
+    }
+    offspring = rank(offspring)
+    return offspring[0]
+}
+
+document.querySelector('#generate').onclick = () => run(20, 20)
+
+document.querySelector('#asex').onclick = () => asexRun(20,20,.7)
```
gigamonkey, line 220: This will be a lot easier to debug if you can see how things are evolving. This code will set things up so you create one critter and then make a new child and keep whichever one is better, in a way that you can see each child be drawn. Once you can see that happening you hopefully will be able to see some other problems, especially if you drop down the number of triangles to 1 or 2.

```javascript
document.querySelector('#asex').onclick = () => {

    let chance = 0.4;
    let best = { 'fitness': 0, 'triangles': randomTriangles() };

    const step = () => {
        const child = batchFitness({ 'fitness': 0, 'triangles': mutate(mix(best.triangles), chance) });
        if (child.fitness > best.fitness) {
            console.log(`child is better`);
            best = child;
        }
        requestAnimationFrame(step);
    };
    requestAnimationFrame(step);
};
```

gigamonkey (changes requested):
You should add in the code I gave you in a comment and then you can debug some other problems.

PR #70, opened 2023-05-11
still doesnt really generate correctly

elijahsur:


gigamonkey (changes requested):
The factor thing is my main suspect. But I can show you in class how I debugged a version of this.

gigamonkey (approved):
You should use the "Rename symbol" in the editor to rename a bunch of those `triangle` variables to `polygon`. Also, it's worth it to be careful about using plurals and singulars correctly in variable names. You had a variable that used to be called `triangles` and you renamed it `polygon`. Which is confusing since it actually represents an array of polygons, not a single one.

PR #71, opened 2023-05-19
Projects/minesweeper

elijahsur:


projects/minesweeper/public/script.js
```diff
@@ -0,0 +1,42 @@
...
+    } else {
+        console.log(`setting value to ${e.target.value}`);
+        size = e.target.value
+    }
+    drawBox()
+};
+
+const drawBox = () => {
+    document.getElementById('grid').innerHTML = "";
+    grid.style.setProperty('gridTemplateColumns',`repeat(${size}, 1fr);`)
```
gigamonkey, line 30: Oh goodness. This is annoying. Turns out the problem is the `;' in the value. Your original plan would have worked fine if you had written this:

```javascript
grid.style.gridTemplateColumns = `repeat(${size}, 1fr)`;
```

With the `;` in the value it's an illegal value and thus has no effect. (The fact that it fails completely silently is pretty annoying.)

The version with`setProperty` can also work but not only do you have to remove the `;`, you need to use the actual name of the property, not the version converted to camelCase. So:

```javascript
grid.style.setProperty('grid-template-columns',`repeat(${size}, 1fr)`);
```

gigamonkey (changes requested):
Found your problem. That was a subtle one.

projects/minesweeper/public/script.js
```diff
@@ -0,0 +1,80 @@
...
+
+const planBombs = () => {
+    let bombArray = [];
+    let bombsLeft = bombcount
+    for (let i = 0; i < size * size; i++) {
+        bombArray.push({i, 'bomb': false})
+    }
+    for (let j = 0; j < bombcount; j++) {
+        bombArray[maxRandom(bombArray.length)].bomb = true
+    }
```
gigamonkey, line 74: I'll note that this might not create `bombcount` bombs if you randomly draw the same random index more than once. There are a number of ways to fix that. What I'd probably do is grab a shuffle routine I've written a billion times, use it to shuffle a copy of the array of objects, and then set the `bomb` property of the first `bombcount` elements  of that shuffled array to `true`. If you want to write your own shuffle routine google "Fisher-Yates shuffle". Or you can probably find the code for that online somewhere or ask ChatGPT. Or there are other ways to solve the problem.

projects/minesweeper/public/script.js
```diff
@@ -0,0 +1,80 @@
...
+
+const randomRGB = () =>
+    Array(3)
+        .fill()
+        .map(() => Math.floor(Math.random() * 255));
+
+const rgb = (r, g, b) => `rgb(${r}, ${g}, ${b})`;
+
+const setColor = (element, location) => {
+    if (location.bomb === true) {
```
gigamonkey, line 16: You should really get used to using booleans for their own value, i.e. not comparing them to literal values.

```javascript
if (location.bomb) {...
```

is totally sufficent. Otherwise why not write:

```javascript
if ((location.bomb === true) === true) {
```

gigamonkey (approved):
Looks like a good start. I like how you're laying down the structure of functions you need and calling them in the right places and leaving actually implementing them to later; that's a great way to build up a program.

